PR #9, opened 2018-10-26
Improved service names

mrnetops:
* -service is now optional, all services pulled by default
* getServiceNames is now one api call instead of one api call per service

### DIFF
--- a/fastly.go
+++ b/fastly.go
@@ -30,7 +30,6 @@ func getServiceNames(token string, serviceIDs []string, logger log.Logger) map[s
 	defer resp.Body.Close()
 
 	var response serviceResponse
-
 	if err := json.NewDecoder(resp.Body).Decode(&response); err != nil {
 		level.Error(logger).Log("err", err)
 		return serviceNames
@@ -42,15 +41,12 @@ func getServiceNames(token string, serviceIDs []string, logger log.Logger) map[s
 
 	if len(serviceIDs) > 0 {
 		limitedServiceNames := map[string]string{}
-
 		for _, serviceID := range serviceIDs {
 			if val, ok := serviceNames[serviceID]; ok {
 				limitedServiceNames[serviceID] = val
 			}
 		}
-
 		return limitedServiceNames
-
 	}
 
 	return serviceNames
@@ -107,8 +103,8 @@ type realtimeResponse struct {
 }
 
 type serviceResponse []struct {
-	ID         string `json:"id"`
-	Name       string `json:"name"`
+	ID   string `json:"id"`
+	Name string `json:"name"`
 }
 
 type datacenter struct {
--- a/fastly.go
+++ b/fastly.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"net/url"
 	"time"
 
 	"github.com/go-kit/kit/log"
@@ -14,17 +13,11 @@ import (
 
 func getServiceNames(token string, serviceIDs []string, logger log.Logger) map[string]string {
 	serviceNames := map[string]string{}
-	for _, serviceID := range serviceIDs {
-		serviceNames[serviceID] = getServiceName(token, serviceID, log.With(logger, "service_id", serviceID))
-	}
-	return serviceNames
-}
 
-func getServiceName(token, serviceID string, logger log.Logger) string {
-	req, err := http.NewRequest("GET", fmt.Sprintf("https://api.fastly.com/service/%s/details", url.QueryEscape(serviceID)), nil)
+	req, err := http.NewRequest("GET", "https://api.fastly.com/service", nil)
 	if err != nil {
 		level.Error(logger).Log("err", err)
-		return serviceID
+		return serviceNames
 	}
 
 	req.Header.Set("Fastly-Key", token)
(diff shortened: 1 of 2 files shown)


fastly.go
```diff
@@ -14,41 +13,47 @@ import (
...
-	return serviceNames
-}
 
-func getServiceName(token, serviceID string, logger log.Logger) string {
-	req, err := http.NewRequest("GET", fmt.Sprintf("https://api.fastly.com/service/%s/details", url.QueryEscape(serviceID)), nil)
+	req, err := http.NewRequest("GET", "https://api.fastly.com/service", nil)
 	if err != nil {
 		level.Error(logger).Log("err", err)
-		return serviceID
+		return serviceNames
```
peterbourgon, line 20: Hmm, this needs more thinking. If -service is optional, and this lookup process fails for whatever reason, the exporter won't do useful work. My intuition is that the program should exit in this case. And I think it's principle-of-least-surprise if the program would exit if getServiceNames failed, regardless if any -service is specified or not. What do you think?

fastly.go
```diff
@@ -14,41 +13,47 @@ import (
...
 	}
 	defer resp.Body.Close()
 
-	var response struct {
-		Name string `json:"name"`
-	}
+	var response serviceResponse
+
```
peterbourgon: Please remove extraneous newline.

fastly.go
```diff
@@ -14,41 +13,47 @@ import (
...
+		limitedServiceNames := map[string]string{}
+
+		for _, serviceID := range serviceIDs {
+			if val, ok := serviceNames[serviceID]; ok {
+				limitedServiceNames[serviceID] = val
+			}
+		}
+
+		return limitedServiceNames
+
```
peterbourgon: Please remove all extraneous newlines in this block.

fastly.go
```diff
@@ -14,41 +13,47 @@ import (
...
+	for _, entry := range response {
+		serviceNames[entry.ID] = entry.Name
 	}
-	if response.Name == "" {
-		level.Error(logger).Log("err", "Fastly returned blank service name")
-		return serviceID
+
+	if len(serviceIDs) > 0 {
+		limitedServiceNames := map[string]string{}
```
peterbourgon, line 43: Maybe it would be nicer if the -service flags didn't flag.Var to a string slice, but rather directly to a map[string]string, where the key is service ID (provided by user) and the value is initially the empty string but could be looked up by this function?  If no -service is given, then this function could do a first pass to look up all service IDs via the API, and kill the process if it doesn't work. I haven't coded it through, but my intuition is that may make things easier—what do you think?

fastly.go
```diff
@@ -101,6 +106,11 @@ type realtimeResponse struct {
 	Error string `json:"error"`
 }
 
+type serviceResponse []struct {
+	ID         string `json:"id"`
+	Name       string `json:"name"`
+}
```
peterbourgon, line 108: Looks like this hasn't been gofmt'd.

main.go
```diff
@@ -49,10 +49,6 @@ func main() {
 		level.Error(logger).Log("err", "-token is required")
 		os.Exit(1)
 	}
-	if len(serviceIDs) <= 0 {
-		level.Error(logger).Log("err", "at least one -service ID is required")
-		os.Exit(1)
-	}
```
peterbourgon, line 55: This is fine, but please denote that -service is optional in the help text, e.g.

```go
fs.Var(&serviceIDs, "service", "Specific Fastly service ID only (optional, repeatable)")
```

peterbourgon (changes requested):
Sure, I like the feature, the implementation just needs a bit more thinking.

fastly.go
mrnetops: This is actually an intermediary in series of patches.

Patch+1: serviceNames maintained in a separate go routine and shared with the rt polling go routines via map & mutex or sync.map.  This addresses service name changes over time, and will also do something like sleep/retry if the /service call or results are not fit.

Patch+2: rt polling go routines actually instantiated and destroyed by the serviceNames go routine.  This addresses services added or removed over time  **this should address your concerns above.**

fastly.go
mrnetops: I'm ambivalent on this one.  I didn't touch the -service flag construction as it works well enough and doesn't impact any features I would like to add ;)

fastly.go
mrnetops: the "getServiceNames fitness & retry or exit" commit should address this in the interim.

main.go
```diff
@@ -49,13 +49,20 @@ func main() {
...
 
 	level.Debug(logger).Log("msg", "looking up service names")
 	serviceNames := getServiceNames(*token, serviceIDs, log.With(logger, "query", "api.fastly.com"))
+	if len(serviceNames) == 0 {
+		level.Error(logger).Log("err", "service name lookup error, retrying")
+		for i := 1; len(serviceNames) == 0; i++ {
```
peterbourgon: Nah, just exit right away. Anyone who cares will be running this thing in a supervisor.

mrnetops:
I think that's everything, let me know.

peterbourgon:
Great, thanks a bunch for this. In reviewing it, I think what you actually want (fully dynamic service lookups, essentially) is best served by a proper, ground-up refactor. I've made a couple of unsatisfying attempts based on your work, but am continuing on, and should have something in a couple days at the most. Hold tight :)

peterbourgon:
These changes have been taken with #10, thank you kindly for your contribution.

PR #10, opened 2018-10-31
Dynamic services

peterbourgon:
Taking #9 as a ground truth, refactor the entire exporter to more natively support dynamic services.

```
            api.fastly.com                             rt.fastly.com
                  |                                          |
             +---------+            +-------+           +---------+
token------->| service |--update--->| name  |<--lookup--| monitor |-+
serviceIDs-->| queryer |            | cache |           |         | |
             |         |            +-------+           |         | |
             |         |           +---------+          |         | |
             |         |--update-->| monitor |--spawn-->|         | |
             |         |           | manager |--kill--->|         | |
             +---------+           +---------+          +---------+ |
                                                          +-|-------+
             +------------+                                 | | update
namespace--->| prometheus |<--------------------------------' |
subsystem--->| metrics    |<----------------------------------'
             +------------+
               |
               v
             +--------+
addr-------->| HTTP   |
             | server |
             +--------+
```


ljagiello:
@peterbourgon you included `v2` binary in PR ;)

peterbourgon:
@ljagiello Ugh, thank you; fixed.

mrnetops:
While /service should control instantiation/deinstantiation of rt monitoring go routines, we might want to leave the door open for additional resolver components (datacenter_id -> datacenter_name, service_id -> service_domains, etc).  Not sure if those should be their own go routines, or added to the main service resolver thread.

mrnetops:
Thinking further, even rt requests probably shouldn't block completely indefinitely, so we can do things like recover gracefully if there is an api.fastly.com outage.  call it a 1 minute timeout or the like.

peterbourgon:
> serviceQueryer probably shouldn't use the default http client

Agree, fixed.

> even rt requests probably shouldn't block completely indefinitely

Agree, fixed.

> While /service should control instantiation/deinstantiation of rt monitoring go routines, we might want to leave the door open for additional resolver components (datacenter_id -> datacenter_name, service_id -> service_domains, etc)

Sure. If and when those feature requests roll in, I can add additional fooQuery → fooCache boxes in the diagram as required.

—

I think this is in a good state, would you mind giving it a review/test?

peterbourgon:
> Default/non-default tags applied to metrics

IMO, out of scope for this tool, until someone proposes a concrete use case.

> TlsvXXTotal metrics should really just be labels for tlsTotal

Mildly against this change; I'd like the tool to have as few opinions as possible on this kind of thing, and have the Prometheus metrics to mirror as much as possible the rt.fastly.com data structures.

> Custom labeling

Ah, a use case :) Can you say more about what you want to stick on things? Would you maintain a fork or would you want to specify labels at runtime?

ljagiello:
> 2. I'm thinking the TlsvXXTotal metics should really just be labels for tlsTotal

Everything below TLS 1.2 should be 0 right now (https://www.fastly.com/blog/requiring-tls-12-fastly-api-control-panel), but I'm personally super curious hows the migration from TLS 1.2 -> TLS 1.3 will be moving. It will not be possible to see that with a single metric.

mrnetops:
For custom labeling, my first tact is something like
Name: [code] [free form name component]
Where the code label is added.  

In our case there is not a 1:1 code <-> service id relationship, so multiple services might use the same code.  ala inventory code for product foo is used for the service ids for related services (subservice 1, sub service 2, etc) and also for the same services across the development lifecycle (dev, qa, staging, prod, etc)

For my specific use case, we could potentially just add a regex transformation to label option to the main branch

-customLabel 'Variable:Match:Transform:Label'
-customLabel 'Name:^\S+:$1:code'  

but that wouldn't help anybody who needed to do say an asset lookup api call.

I'm not actually sure the best way to maintain a custom addition, I'm open to suggestions.

For the TlsvXXTotal metrics, I would counter that we already perform an equivalent manipulation (appropriately in my opinion) with status_group_total and status_code_total

@ljagiello for your use case, I would say we should be able to trivially graph tlsTotal {version=*} to see what you're interested (I'm totally interested in graphing that delta as well ;) )

peterbourgon:
@mrnetops Isn't this the job of Prometheus relabeling? It feels strictly out of scope for a simple third-party exporter, but as the meme goes, _change my mind_... ;)

mrnetops:
@peterbourgon oooh, I think you're exactly right.  I was not aware that functionality was available on the prometheus side.

I'll rescind my custom label use case while I go play with prometheus relabeling.

I still think my tls comments stand given that we're already doing bits like

https://github.com/peterbourgon/fastly-exporter/blob/a24e11f947df99006e0ec18efdd3b9979e695949/process.go#L64-L68

peterbourgon:
I agree with you on TLS and will make that change.

peterbourgon:
I'll merge this and release v2.0.0, unless you have remaining objections or must-have changes.

mrnetops:
Looks fantastic to me sir, this is awesome!



Per your suggestion, working prometheus metric_relabel_configs example for my use case 
```yaml
  - job_name: 'fastly-exporter'

    # metrics_path defaults to '/metrics'
    # scheme defaults to 'http'.
    metrics_path: '/metrics'

    static_configs:
    - targets: ['localhost:8080']

    metric_relabel_configs:
    - source_labels: [service_name]
      regex: '^(\S+)'
      replacement: '$1'
      target_label: inventory_code
```

PR #11, opened 2018-11-06
Minimal Dockerfile

mrnetops:
Tiny little docker images

```text
docker images fastly-exporter
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
fastly-exporter     latest              9fbf45d3fe99        7 minutes ago       11.2MB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,25 @@
+# Based off https://medium.com/@chemidy/create-the-smallest-and-secured-golang-docker-image-based-on-scratch-4752223b7324
+
+# STEP 1 build executable binary
+FROM golang:alpine as builder
+# Install git + SSL ca certificates
+RUN apk update && apk add git && apk add ca-certificates
+# Create appuser
+RUN adduser -D -g '' appuser
+COPY . $GOPATH/src/mypackage/myapp/
+WORKDIR $GOPATH/src/mypackage/myapp/
+#get dependancies
+RUN go get -d -v
+#build the binary
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -installsuffix cgo -o /go/bin/fastly-exporter
+# STEP 2 build a small image
+# start from scratch
+FROM scratch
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,25 @@
+# Based off https://medium.com/@chemidy/create-the-smallest-and-secured-golang-docker-image-based-on-scratch-4752223b7324
+
+# STEP 1 build executable binary
+FROM golang:alpine as builder
```
ljagiello, line 4: With alpine version like: `1.11.2-alpine` it will be possible to control what Go version will be use for a build.

mrnetops:
@peterbourgon Looking over previous pull requests, I noticed that you had closed previous (simpler ;) ) dockerfile merge requests.

I'd like to take another swing at convincing you.

Ideally we're all contributing these features back to the main codebase so we're not individually running into the same issues over and over again.

- 20x larger containers because a single stage rather then multi stage build was used
- Gotchas exposing the docker image ports because fastly-exporter binds to 127.0.0.1 instead of 0.0.0.0 by default
- SSL certificate validation issues for api.fastly.com calls unless a ca-bundle is included

Given at least three dockerfile merge requests so far, I would suggest the userbase is really really interested in having a pre-built containerization option.  I would also suggest that this will not be the last Dockerfile merge request you will receive ;)

For my particular use case I need the fastly-exporter containerized so I can run it in kubernetes.  I can maintain a separate Dockerfile and publish to my private docker repository, but in the spirit of this open project, I'd much prefer not only to get this (or an equivalent) Dockerfile added, but also to get fastly-exporter docker images pushed out directly to the public docker hub as part of the release / continuous integration process.

peterbourgon:
Accepting a PR means taking ownership for maintaining that bit of code in perpetuity. I don't use Docker, so I have no way of responsibly maintaining a Dockerfile. Sorry! 😬

peterbourgon:
With that said, if someone wants to take the trouble of building Docker images and submitting them to Docker Hub, I'm happy to link to them in the README or whatever as the semi/official Docker solution.

PR #18, opened 2018-11-09
Read token from environment variable

ljagiello:
New option: read environment variable `FASTLY_API_TOKEN`.

With this patch `fastly-exporter` will first check `-token` and if it's not there will try read environment variable `FASTLY_API_TOKEN`.

For security reasons we don't want to always leave token as an argument. It also simplifies setup in Kubernetes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ VERSION
 ```
 
 A valid Fastly API `-token` is mandatory. [See this link][token] for information
-on creating API tokens. Token can be provided via environment variable
+on creating API tokens. The token can also be provided via the environment variable
 `FASTLY_API_TOKEN`.
 
 Optional `-service` IDs can be specified to limit monitoring to specific
--- a/README.md
+++ b/README.md
@@ -56,10 +56,10 @@ VERSION
 ```
 
 A valid Fastly API `-token` is mandatory. [See this link][token] for information
-on creating API tokens. 
+on creating API tokens. Token can be provided via environment variable `FASTLY_API_TOKEN`.
 
 Optional `-service` IDs can be specified to limit monitoring to specific
-services. Service IDs are available at the top of your [Fastly dashboard][db]. 
+services. Service IDs are available at the top of your [Fastly dashboard][db].
 
 [token]: https://docs.fastly.com/guides/account-management-and-security/using-api-tokens#creating-api-tokens
 [db]: https://manage.fastly.com/services/all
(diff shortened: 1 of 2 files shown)


mrnetops:
I like this!

ljagiello:
@msuterski It's exactly what I'm trying to avoid ;) I don't want to leave API key in process list (visible to everyone via a simple `ps`).

In my case at least API key is fetch from Vault via init container and pass to main container as a default environment variable. If I could read that without leaving API key in process list it would be super nice.

On top of that in you example you could just skip entire `args` part.

README.md
```diff
@@ -56,10 +56,10 @@ VERSION
 ```
 
 A valid Fastly API `-token` is mandatory. [See this link][token] for information
-on creating API tokens. 
+on creating API tokens. Token can be provided via environment variable `FASTLY_API_TOKEN`.
```
peterbourgon: > The token can also be provided via the `FASTLY_API_TOKEN` environment variable.

And please wrap at 80col.

main.go
```diff
@@ -45,8 +45,10 @@ func main() {
 	}
 
 	if *token == "" {
-		level.Error(logger).Log("err", "-token is required")
-		os.Exit(1)
+		if *token = os.Getenv("FASTLY_API_TOKEN"); *token == "" {
+			level.Error(logger).Log("err", "-token or FASTLY_API_TOKEN is required")
```
peterbourgon, line 49: Please also amend the help text for the -token flag to e.g.

```go
token = fs.String("token", "", "Fastly API token (required; also via FASTLY_API_TOKEN)")
```

peterbourgon (approved):
Yep, sure. A couple of little nits but LGTM.

main.go
ljagiello: Fixed in https://github.com/peterbourgon/fastly-exporter/pull/18/commits/29b472ab3c90bc5d05b87a269f7b8d25338856e1

README.md
ljagiello: Fixed in https://github.com/peterbourgon/fastly-exporter/pull/18/commits/29b472ab3c90bc5d05b87a269f7b8d25338856e1

README.md
```diff
@@ -56,10 +56,11 @@ VERSION
 ```
 
 A valid Fastly API `-token` is mandatory. [See this link][token] for information
-on creating API tokens. 
+on creating API tokens. Token can be provided via environment variable
```
peterbourgon:
```suggestion
on creating API tokens. The token can also be provided via the environment variable
```

Issue #16, opened 2018-11-08
fastly_rt_service_info for group_left/group_right metadata


mrnetops:
https://www.robustperception.io/exposing-the-software-version-to-prometheus
https://www.robustperception.io/how-to-have-labels-for-machine-roles

Technically, we could do all kinds of fun stuff like
* fastly_rt_service_info
* fastly_rt_datacenter_info
* fastly_rt_domain_info
* fastly_rt_customer_info

and stuff all sorts of other one off tidbits in without binding them to the individual metrics.

We could even do service_name that way

something like
```
fastly_rt_service_info{service_id="$SID", service_name="$NAME", customer_id="$CID"} 1
fastly_rt_datacenter_info{datacenter_code="$CODE", datacenter_name="$NAME", datacenter_group="$GROUP", datacenter_shield="$SHIELD"} 1
```

peterbourgon:
No objection, modulo the implementation. I suspect this should be very minimal changes.

peterbourgon:
So we now have service_info with labels service_id, service_name, and service_version. In theory we could labels there for anything [returned by /service](https://docs.fastly.com/api/config#service), but I want to be judicious. Is customer_id really valuable in this context? Are there other fields that have clear and obvious value?

For datacenter, domain, and customer, I think it's a little less clear. These are meta-properties of your configuration or even Fastly itself, and maybe not directly related to individual services. I'm open to adding additional stuff like that, but I'd like to understand the use cases that information would enable. Do you have any? :)

peterbourgon:
#80 adds the datacenter_info metric and establishes a pattern for arbitrary metadata. This issue is a bit general, so I'm gonna go ahead and close it — please open new issues for any additional metadata you'd like to see.

PR #19, opened 2018-11-10
Compose

mrnetops:
Docker Compose for a full fastly-exporter + Prometheus + Grafana + Fastly dashboard stack

```
FASTLY_API_TOKEN=${FASTLY_API_TOKEN} docker-compose up
```
Access the Grafana dashboard via http://localhost:3000

fastly-dashboard


README.md
```diff
@@ -63,3 +63,24 @@ services. Service IDs are available at the top of your [Fastly dashboard][db].
...
+[Docker Compose][compose] for a full fastly-exporter + [Prometheus][prom] + 
+[Grafana][grafana] + Fastly dashboard stack
+
+[compose]: https://github.com/docker/compose
+[grafana]: https://grafana.com
+
+```
+FASTLY_API_TOKEN=${FASTLY_API_TOKEN} docker-compose up
+```
+Access the [Grafana][grafana] dashboard via http://localhost:3000
```
peterbourgon:
```suggestion

Access the [Grafana][grafana] dashboard via http://localhost:3000.
```

README.md
```diff
@@ -63,3 +63,24 @@ services. Service IDs are available at the top of your [Fastly dashboard][db].
...
+
+![Fastly Dashboard in Grafana](https://github.com/peterbourgon/fastly-exporter/compose/Fastly-Dashboard.png)
```
peterbourgon: Grotesquely pedantic nit: can you make this e.g. `dashboard.png` :)

docker-compose.yml
```diff
@@ -0,0 +1,40 @@
...
+      - prometheus
+    ports:
+      - 3000:3000
+    volumes:
+      - ./compose/grafana/provisioning/:/etc/grafana/provisioning/
+      - grafana_data:/var/lib/grafana
+
+  fastly-exporter:
+    image: mrnetops/fastly-exporter
+    command: -token ${FASTLY_API_TOKEN}
```
peterbourgon: You want to leverage the upcoming env var to avoid sticking the token in `ps` output?

peterbourgon:
re: origin latency, isn't that [miss_duration_seconds](https://github.com/peterbourgon/fastly-exporter/blob/2e922a7505dd474cd1e03c6fb2b91f42727536ac/prometheus_metrics.go#L345)?

re: global POP traffic, I'm not sure exactly what you're after; can you say more? There is the synthetic "aggregated" datacenter, explained thusly:

> The keys of the datacenter JSON object are the names of each of our POPs around the world plus a special key, called aggregated, which adds all the values from all the POPs together.

The fastly-exporter collects that data, but doesn't do anything with it, because in Prometheus you can just sum over the datacenter dimension and get the same aggregation at query time.

docker-compose.yml
mrnetops: I'll adjust it to pure ENV when @ljagiello finishes their edits and we merge.  I suspect he'll be done before we finish out the review.

I know you're not a docker (or I assume docker-compose) fan but you should check this out, it's hella cool to see in action.

mrnetops:
re: global POP traffic that just ties back into #16  I need the region or latitude / longitude to place stats on the map.  i.e. rt.fastly.com + api.fastly.com has the data.  I'm pretty sure I just need to get the data into fastly_rt_datacenter_info so I can pivot off of it.

peterbourgon:
Oh I see it's going to be a map viz. Sure, do you know which Fastly API has that stuff? It should be easy to one-off it at program start. (Certainly not worth pinging the API every 5 minutes to see if POP locations have changed ;))

mrnetops:
Aye! https://docs.fastly.com/api/tools#datacenter to populate fastly_rt_datacenter_info

I'm trying to replicate the standard fastly service dashboard in it's entirety for my initial canned dashboard.

mrnetops:
re: origin latency.  ooooh, you may be onto something there!

Also, I'd love some thoughts on `HIT TIME` and `MISS TIME`, I'm having a hard time getting my dashboard to line up with the stock fastly dashboard for that.  `HIT TIME` seems to be an order of magnitude off, and `MISS TIME` is just weird.

peterbourgon:
What is "the stock Fastly dashboard"?

mrnetops:
https://manage.fastly.com/dashboard/services/<service_id>/datacenters/all

docker-compose.yml
ljagiello: @mrnetops it's done now.

mrnetops: Docker hub updated (@peterbourgon we probably want to tag and release after this PR) and update added!

Nice addition and suggestion @ljagiello @peterbourgon

mrnetops:
@peterbourgon let me know if you have any last edits, otherwise I'll merge the PR

peterbourgon (approved):

PR #27, opened 2019-07-02
Add flag -pollfreq

kkuehlz:
Let the user specify how often to poll the Fastly API. Defaults to 1
minute for backwards backwards compatibility.

### DIFF
--- a/main.go
+++ b/main.go
@@ -28,6 +28,7 @@ func main() {
 		addr       = fs.String("endpoint", "http://127.0.0.1:8080/metrics", "Prometheus /metrics endpoint")
 		namespace  = fs.String("namespace", "fastly", "Prometheus namespace")
 		subsystem  = fs.String("subsystem", "rt", "Prometheus subsystem")
+		pollfreq   = fs.Duration("pollfreq", time.Minute, "How often to poll Fasty API")
 		debug      = fs.Bool("debug", false, "Log debug information")
 	)
 	fs.Var(&serviceIDs, "service", "Specific Fastly service ID (optional, repeatable)")
@@ -99,11 +100,11 @@ func main() {
 
 	var g run.Group
 	{
-		// Every minute, query Fastly for new services and their names.
+		// Query Fastly for new services and their names.
 		// Update our name cache and managed monitors accordingly.
 		var (
 			ctx, cancel = context.WithCancel(context.Background())
-			ticker      = time.NewTicker(time.Minute)
+			ticker      = time.NewTicker(*pollfreq)
 		)
 		g.Add(func() error {
 			for {


peterbourgon:
I'm struggling to think of a use case where this default would need to be changed. Can you tell me yours?

kkuehlz:
@peterbourgon On our dev server we are running a bunch of different exporters, and the fastly exporter uses a bunch of resources at pulls in a lot of data, so I'd like the frequency that data gets pulled in to be configurable, if you are open to it. Similar to `period_seconds` in `cloudwatch-exporter`.

peterbourgon:
To be longer or shorter? And you're trying to do this to reduce resource consumption of the exporter process itself? Can you say more about that?

~The thing this bumps against is the design of rt.fastly.com. Every time you make a request, it will block until the next of its internal epochs is reached, and then return aggregate data for that window. Changing the interval with which we poll won't have any real effect, except maybe reducing the granularity of the data exposed to Prometheus. There may be some wiggle room on this, I'll ask the developer about that and report back...~ _edit_: I had a brain freeze, see below comment.

peterbourgon:
Oh, gosh, sorry, I've totally mis-read what this bit of the code does. This bit polls api.fastly.com for new services associated with your token (if no service IDs were explicitly passed) and/or updated service names for those services. It is—or ought to be—a low-bandwidth, low-resource operation. It's totally disconnected from hitting the main rt.fastly.com real-time stats API.

Is this still what you intended to tweak? If so, can you say in more detail why you'd want to tweak it?

kkuehlz:
@peterbourgon Ah, no worries. I misinterpreted the code too. For what I wanted, I think I should've added code to the loop in `monitor.go` that is hitting rt.fastly.com. Would you be open to this?

In short, I want the rate at which we poll the fastly api for metrics to be configurable , so I think your above comment is relevant.

peterbourgon:
> For what I wanted, I think I should've added a larger sleep period to the loop in monitor.go that is hitting rt.fastly.com. Would you be open to this?

OK, so we can talk a bit more about why this is probably pointless :)

The way rt.fastly.com works is you make a request with a service ID and a timestamp, and it blocks while it collects real-time counters for your service from that timestamp forward. If your service is reasonably high-traffic, it will return after about a second. If your service isn't so high-traffic, it will block for up to 30 seconds before returning. In all cases, the response represents traffic for your service, for the time window (request timestamp -> response timestamp). So you're meant to call it in a hot loop, passing the response timestamp of iteration _n_ as the request timestamp of iteration _n+1_.

As far as I can tell, adding a delay to this hot loop won't do anything meaningful. If you delay more than a minute between calls you'll probably just lose data, as rt.fastly.com doesn't keep a lot of history. In that case you might look at writing an exporter for the [historical stats API](https://docs.fastly.com/api/stats), though I'm not sure how much sense that would make. If you delay for a handful of seconds between calls, you might make fewer calls (for high traffic services) but the data you get back will be the same, just with lower granularity.

So, again: what specifically are you trying to solve with this change? If fastly-exporter is running too hot in your environment, can you say more about that? Is it burning too much CPU? Is it using too much memory? Maybe we can solve it via another means.

kkuehlz:
@peterbourgon Thanks for the detailed response. It is mostly CPU. It is usually at 10-11% utilization, and every 10 seconds or so spikes and pins a core. Since we bumped the specs on our machine it's not a _huge_ issue. I mistakenly thought this would be a quick fix.

peterbourgon:
If you were interested enough, I could walk you through taking a profile, and we could see what was burning the CPU. My intuition is that it's JSON deserialization, how many services are you collecting stats for? But if you don't want to bother, I totally understand.

kkuehlz:
@peterbourgon When we put this exporter in prod, this won't be a problem for us, so it's not critical. Though if it will help out your project, I'm happy to help profile it. We have 85 services that we are collecting stats from, and a lot of domains.

peterbourgon:
@keur

> We have 85 services that we are collecting stats from, and a lot of domains.

Oh wow, that might do it :)

I've pushed a `pprof` branch with profiling enabled. If you want to help me out, please build that and drop it in your test environment, get it running, and then take a profile. Something like this should work:

```
$ go get github.com/peterbourgon/fastly-exporter
$ cd $GOPATH/src/github.com/peterbourgon/fastly-exporter
$ git checkout pprof
$ go build .
$ ./fastly-exporter <your options here>

$ go tool pprof -svg http://localhost:8080/debug/pprof/profile ./fastly-exporter
```

And then attach the resulting SVG, or, if you'd rather not make it public, send it via email to my first initial and my last name at Fastly dot com. Any problems, please let me know. And thanks again for offering!

peterbourgon:
Profiles were emailed and confirm it's mostly JSON unmarshaling. We have a couple of options to make that less CPU intense, I'll explore which ones are the most elegant.

Issue #28, opened 2019-07-11
Uncontrolled memory usage growth when covering a large number of service


dvhh:
covering >20 services ( using the docker container ), the memory usage is increasing steadily until the memory limit is reached (~1GB)
![image](https://user-images.githubusercontent.com/573842/61054047-b2175a00-a429-11e9-8cff-78a1d6eb6ebe.png)

peterbourgon:
Which memory metric is this?

peterbourgon:
Can you give https://github.com/peterbourgon/fastly-exporter/releases/tag/v2.2.1 a shot?

dvhh:
Hello, Sorry for the late reply, will give the version test and will report if same issue occur, unfortunately the Memory metric is an ad-hoc monitoring expressing memory usage in % of the container ( an probably missing a few beats ).

![image](https://user-images.githubusercontent.com/573842/61101100-9486d680-a4a3-11e9-9d8e-262ade17ba29.png)

metrics for :
- go_memstats_heap_alloc_bytes (stable)
- go_memstats_heap_idle_bytes (increase during lifetime)
- go_memstats_heap_sys_bytes

These metrics are produced before using the suggested version

dvhh:
Started running the new version, will update with reported metrics in ~24h

peterbourgon:
Bear in mind also that [the operating system may not necessarily reclaim idle memory](https://github.com/golang/go/issues/14521) and thus steadily increasing values there don't necessarily imply a memory leak.

dvhh:
![image](https://user-images.githubusercontent.com/573842/61168090-0124d280-a583-11e9-91e5-1caf24f1bbe8.png)
current status

peterbourgon:
Cool, I’ll call this fixed unless you object.

dvhh:
No issue, the application seems to be more stable in its memory consumption.
Feel free to close the ticket.

dvhh:
Just to update on the status
based on limits for killing containers (limited to 1GB) 
- version 2.2.0 : ~30h uptime for a container
- version 2.2.1 : ~34h uptime for a container
Thanks for the support, and the clarification about where the issue mainly remain

dvhh:
version 3.0.1: uptime ~5 days and ongoing, memory usage stable, thanks for your work

peterbourgon:
Great news, thanks for the reports.

PR #29, opened 2019-07-11
Fix potential leak, use modules, various cleanups

peterbourgon:
Just basic maintenance.


peterbourgon:
Might help address #28.

PR #39, opened 2020-01-13
enable exclusion of measurements as cli arg

j18e:
this feature will allow for the explicit exclusion of measurements using command line args. Have tested thoroughly.

### DIFF
--- a/cmd/fastly-exporter/main.go
+++ b/cmd/fastly-exporter/main.go
@@ -30,21 +30,23 @@ var programVersion = "dev"
 func main() {
 	fs := flag.NewFlagSet("fastly-exporter", flag.ExitOnError)
 	var (
-		token       = fs.String("token", "", "Fastly API token (required; also via FASTLY_API_TOKEN)")
-		addr        = fs.String("endpoint", "http://127.0.0.1:8080/metrics", "Prometheus /metrics endpoint")
-		namespace   = fs.String("namespace", "fastly", "Prometheus namespace")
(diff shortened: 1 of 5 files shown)


peterbourgon:
Can you give a brief description of your use case?

I'm OK with this feature, but I would implement it differently :) I'll happily do that work once I understand what you're aiming for.

j18e:
we're using this at finn.no (in Norway) and it's returning over 60,000 time series. We're trying to cut down on this and the main way we wish to do this is excluding metrics we don't actually use from scrapes.

I know this is achievable through the scrape configs in prometheus itself, but this results in somewhat of a scattered configuration, which we're trying to avoid.

Would love to hear how you'd implement it. I tried to do it with minimal code changes but I'm sure it could have been done better :)

Thanks for the quick response!

peterbourgon:
Would it make sense to have this work like the whitelist and blacklist flags for service names, except applied to metric names?

j18e:
Absolutely! I'll work on it and get back to you

peterbourgon:
Oh, I'm happy to do it — it will trigger some other refactors. Are you OK with using a single regex string to {whitelist, blacklist} the metrics you {want, don't want} — or would you strongly prefer to repeat flags?

peterbourgon:
After a bit of thought, I think I can make the -include-regex and -exclude-regex flags repeatable in all cases.

j18e:
Okay that sounds good. Thanks a bunch!

Jamie Wiebe

peterbourgon:
Solved with #40, thanks for the use case.

PR #40, opened 2020-01-13
Allow filtering of exported metrics

peterbourgon:
This PR is a re-work of #39 that allows you to filter the Prometheus metrics that get exported. It is a **breaking change** in order to make the flags for service and metric name filtering consistent: `-{service, metric}-{whitelist, blacklist}`. The full semantics of the filtering behavior are explained in the README.

@j18e — does this solve your use case? If so, I'll push a new release shortly.


j18e:
works perfectly on my end

Issue #35, opened 2019-07-25
Support rt.fastly.com "demo" channel


vannahj28:
```
 ./fastly-exporter-3.0.1-linux-amd64 -token xyz -service demo
level=info prometheus_addr=127.0.0.1:8080 path=/metrics namespace=fastly subsystem=rt
level=info component=api.fastly.com filtering_on="explicit service IDs" count=1
level=error component=api.fastly.com during="initial service refresh" err="error decoding API services response: json: cannot unmarshal object into Go value of type []api.Service"

```

There is a special channel ID demo which is used for the analytics on the fastly.com
(https://www.fastly.com) homepage.

Source: https://docs.fastly.com/api/analytics#channels

Useful for populating test data

peterbourgon:
Supporting the special-case `demo` channel ID on rt.fastly.com would require a lot of special-casing throughout the fastly-exporter: it must bypass the api.fastly.com lookup stages, ignore any provided -token, etc. And I'm not even sure I want to allow it in the first place, as it can generate a lot of load. Still, I'll think if there's a reasonable way to get there.

PR #42, opened 2020-01-14
Add -demo flag to export fastly.com metrics

peterbourgon:
#35


peterbourgon:
Not planning to do this at the moment, and the PR bit-rotted.

PR #45, opened 2020-03-23
Add additional environment variable options

dgarlitt:
Add ability to dynamically set some additional start-up options via environment variables for docker images.

New options:
 - PROMETHEUS_METRICS_ENDPOINT
 - FASTLY_SERVICE_IDS


peterbourgon:
If it's OK with you I'll just switch to [ff](https://github.com/peterbourgon/ff) for flags parsing, and make all of the flags available as env vars with FASTLY_EXPORTER_ as a prefix. Sound OK?

PR #46, opened 2020-03-23
Use package ff, enabling env vars and config files

peterbourgon:
Alternative to #45. @dgarlitt, does this serve the purpose?

### DIFF
--- a/cmd/fastly-exporter/main.go
+++ b/cmd/fastly-exporter/main.go
@@ -11,6 +11,7 @@ import (
 	"os/signal"
 	"strconv"
 	"strings"
+	"text/tabwriter"
 	"time"
 
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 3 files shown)


peterbourgon:
Oops, you also want FASTLY_EXPORTER_SERVICE_IDS split on commas, eh?

mrnetops:
@peterbourgon we probably want that for all fields that can potentially do multiple values, whitelist, blacklist etc.  

I like that you just tossed in a potential config file too ;)

peterbourgon:
Unfortunately the regexp fields can contain commas. Is it smart to allow those to be provided via env vars if they're not a complete solution?

mrnetops:
Hmmm.  Initial less clean thoughts

0) Is there a reasonable way to support bash arrays in golang?
1) regexp from environment variables only get one value, but being regexps, you can OR them together.  i.e. "this|that" vs "this,that"
2) less clean, given that ',' is likely only going to be used for -service, -service-whitelist and -service-blacklist we could include the addendum that "," is a hard separator and inline commas would need to be replaced by "."

...I did say less clean thoughts ;)

peterbourgon:
@dgarlitt Could you say a bit about why you want more flags controllable by env vars? Is it not possible to pass flags in your runtime environment?

PR #57, opened 2020-10-22
Use correct API field name for edge request count

tomhughes:
The edge request count is always showing as zero because it's mapped to the wrong API field name...


peterbourgon:
Huh, I thought I wrote a test for this...

peterbourgon:
Nope, just a typo. Thanks!

PR #58, opened 2020-10-22
Avoid double counting total requests

tomhughes:
Total requests are already counted by the defined mapping so there's no need to count them manually as well.


peterbourgon:
Ah, thanks! Another brainfart.

PR #70, opened 2021-08-26
Add basic http status pages.

SuperQ:
* Add a / landing page for the exporter.
* Add a /-/ready endpoint for supervisors.

Signed-off-by: SuperQ <superq@gmail.com>

### DIFF
--- a/cmd/fastly-exporter/main.go
+++ b/cmd/fastly-exporter/main.go
@@ -319,8 +319,24 @@ func main() {
 		})
 	}
 	{
+		// Serve landing page.
+		http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+			w.Write([]byte(`<html>
+             <head><title>Fastly Exporter</title></head>
+             <body>
+             <h1>Fastly Exporter` + programVersion + `</h1>
+             <p><a href='` + promURL.Path + `'>Metrics</a></p>
+             </dl>
+             </body>
+             </html>`))
+		})
 		// Serve Prometheus metrics (and /debug/pprof/...) over HTTP.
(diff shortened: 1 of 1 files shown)


peterbourgon:
> Add a `/` landing page...

Mmmmmaybe. Is there a concrete problem being solved with this?

> Add a `/-/ready` endpoint for supervisors

Anything which needs to know the health of the exporter can `HEAD` the metrics endpoint. See #61. If there's something I'm overlooking, though, please tell me.

SuperQ:
It's pretty standard for exporters to have a root landing page, rather than the current behavior of 404.

[Kubernetes standard readiness probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) do a GET. There is no option to change the request method.

Doing a GET on the metrics endpoint can be very large, and Kubernetes will probe this regularly, eating up a lot of resources in both the exporter and the Kubelet services.

peterbourgon:
> Doing a GET on the metrics endpoint can be very large

Agreed, this isn't a solution, but...

> There is no option to change the request method.

This is broken 🤷

peterbourgon:
> It's pretty standard for exporters to have a root landing page

Sure, but who's going there? Humans, when debugging or whatever?

SuperQ:
Yes, the root page is a convenience page for humans.

Also, this is just a first pass. Eventually we could tie the readiness check into if the exporter has a valid connection to the RT API. This way Kubernetes can avoid sending traffic (Prometheus) to a broken instance. It also informs Kubernetes to wait to tear down the old instance(s) when doing upgrades.

SuperQ:
Closing in favor of #73

Issue #66, opened 2021-07-22
Add discovery/scrape-time service selection


SuperQ:
With Prometheus 2.28, there is now a [generic http service discovery](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#http_sd_config).

The exporter can now produce an API output that lists all of the available services so that they can be scraped independently. This improves the performance of ingestion by spreading it out over time and allowing Prometheus to ingest the data over multiple target threads.

On the Prometheus side, you would configure the job like this:

```yaml
scrape_configs:
- job_name: fastly
  metrics_path: /fastly
  http_sd_configs:
  - url: http://fastly-exporter:8080/sd
  relabel_configs:
  - source_labels: [__address__]
    target_label: __param_target
  - source_labels: [__param_target]
    target_label: instance
  - target_label: __address__
    replacement: fastly-exporer:8080 
```

The `/service-discovery` endpoint would output json like this:
```json
[
  { 
    "targets": [
      "<Service ID 1>",
      "<Service ID 2>",
      "<Service ID 3>",
      "<Service ID ...>"
    ]
  }
]
```

The `relabel_config` would then produce exporter URLs like `/fastly?service=<Service ID 1>`.

SuperQ:
This is meant as an additional improvement on https://github.com/peterbourgon/fastly-exporter/issues/31. Possibly a replacement, depending on where the bottlenecks are.

CC @neufeldtech

peterbourgon:
Super cool.

peterbourgon:
So wait, given `/discovery` returns

```json
[
  { "targets": ["A", "B", "C"] }
]
```

there's no absolute or explicit URL mapping to each of A, B, and C — I would just assert in the exporter (or a flag or something) that they map to e.g. `/service/{A, B, C}` which the scraping Prometheus would have to be configured to match a priori. Is that right?

SuperQ:
Yes, it needs to be re-mapped by a relabel config in Prometheus no matter what. The discovery injects the target list items into the `__address__` meta-label in discovery. Prometheus doesn't treat those as URLs, it treats them as `host:port` pairs.

peterbourgon:
[...]

PR #73, opened 2021-09-12
Refactor for Prometheus HTTP service discovery

peterbourgon:
This PR will be a refactor to address #66 and part of #70.

- `-endpoint <url>` removed, `-listen [<host>]:<port>` added
- `/metrics` still yields metrics for all services
- `/metrics?target={serviceID}` now does a single service
- `/sd` should be compatible with [Prometheus generic HTTP SD](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#http_sd_config)
- Add index page

### DIFF
--- a/pkg/prom/registry.go
+++ b/pkg/prom/registry.go
@@ -1,0 +1,205 @@
+package prom
+
+import (
+	"encoding/json"
+	"fmt"
+	"html/template"
+	"net/http"
+	"net/url"
+	"sort"
+	"strings"
+	"sync"
+
+	"github.com/gorilla/mux"
+	"github.com/peterbourgon/fastly-exporter/pkg/filter"
+	"github.com/peterbourgon/fastly-exporter/pkg/gen"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+// Registry collects Prometheus metrics on a per-service basis.
+//
+// Writers (i.e. rt.Subscribers) should call MetricsFor with their specific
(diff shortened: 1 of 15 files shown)


peterbourgon:
@SuperQ Is this what you had in mind? Also, I'd like to provide an example Prometheus HTTP SD config in the README that would make this work, could you help with that?

pkg/prom/registry.go
```diff
@@ -0,0 +1,213 @@
...
+		metricNameFilter: metricNameFilter,
+		byServiceID:      map[string]*metricsRegistry{},
+	}
+
+	router := mux.NewRouter()
+	router.StrictSlash(true)
+	router.Methods("GET").Path("/").HandlerFunc(r.handleIndex)
+	router.Methods("GET").Path("/sd").HandlerFunc(r.handleServiceDiscovery)
+	router.Methods("GET").Path("/metrics").HandlerFunc(r.handleMetrics)
+	router.Methods("GET").Path("/metrics/{serviceID:.+}").HandlerFunc(r.handleMetricsFor)
```
SuperQ: To make this better follow the Prometheus conventions, we use a `__param_target` in the relabel config to send this via URL param, rather than change the `metrics_path` of the service discovery.

Something like this:
```suggestion
	router.Methods("GET").Path("/metrics").Queries("target", "{serviceID:.+}").HandlerFunc(r.handleMetricsFor)
	router.Methods("GET").Path("/metrics").HandlerFunc(r.handleMetrics)
```

SuperQ (approved):
LGTM, Nice

PR #75, opened 2021-09-26
Publish Docker images

peterbourgon:
This PR automates the release process, including creating a Docker image in ghcr.io, whenever a new tag is pushed.

@mrnetops Does this Docker image seem OK to you?


mrnetops:
probably going to want the app to not run as root in the docker image with

# Create appuser
RUN adduser -D -g '' appuser

USER appuser

peterbourgon:
~Is that still relevant given `FROM scratch`?~

_edit_: Apparently it is good practice.

PR #76, opened 2021-09-27
Update README with example http_sd_config

peterbourgon:
@SuperQ This seems to work for me locally, does it pass your smell test?


SuperQ (approved):
LGTM

PR #80, opened 2021-11-16
Add datacenter_info metric

### LINKED ISSUE #79: Add datcenters API enricment
There's a [`/datacenters`](https://developer.fastly.com/reference/api/utils/pops/#list-pops) API endpoint that has a bunch of additional info that could be useful as labels.

For example, the `group` might be a nice label to aggregate over.

peterbourgon:
Contains limited metadata about datacenters. Closes #79.


pkg/api/datacenter_cache.go
```diff
@@ -0,0 +1,113 @@
...
+
+type DatacenterCache struct {
+	client HTTPClient
+	token  string
+
+	mtx sync.Mutex
+	dcs []datacenter
+}
+
+type datacenter struct {
```
mrnetops: latitude and longitude would be nice for doing map based dashboards, ala https://grafana.com/grafana/plugins/grafana-worldmap-panel/ or the like

mrnetops: hmm, highlighting didn't work how I had hoped, meant to highlight the full datacenter struct 3 lines down, but hopefully still makes sense ;)

SuperQ: I forget how that plugin works. Does it need separate lat/on metrics as values? Or can it use label strings as the values?

mrnetops: I'm guessing values, probably directly as latitude & longitude

peterbourgon:
If it's values, that means adding two new static timeseries per datacenter. That doesn't seem like a cost worth the benefit. It's gotta be possible with labels, right?

pkg/api/datacenter_cache.go
peterbourgon: If it's values, that means adding two new static timeseries per datacenter. That doesn't seem like a cost worth the benefit. It's gotta be possible with labels, right?

mrnetops: Gah!  I believe you're right, and I suspect I might have transposed that in my head.

peterbourgon: Cool, this basically works out-of-the-box with the geomap panel, just set format to table.

Screen Shot 2021-11-18 at 4 19 00 PM

pkg/api/datacenter_cache.go
```diff
@@ -0,0 +1,117 @@
...
+	dcs := make([]Datacenter, len(c.dcs))
+	copy(dcs, c.dcs)
+	return dcs
+}
+
+// Gatherer returns a Prometheus gatherer which will yield current metadata
+// about Fastly datacenters as labels on a gauge metric.
+func (c *DatacenterCache) Gatherer(namespace, subsystem string) (prometheus.Gatherer, error) {
+	var (
+		fqName      = prometheus.BuildFQName(namespace, subsystem, "datacenter_info")
```
SuperQ, line 88: Hmm, the `subsystem` makes this `fastly_rt`, rather than `fastly_api`.

pkg/api/datacenter_cache.go
SuperQ: Awesome!

pkg/api/datacenter_cache.go
peterbourgon: My view is that the "subsystem" of this exporter is Fastly real-time stats, and the fact that we also include some metadata from the normal API is kind of incidental. I admit in this specific case it's a bit of a gray area since the metric deals only with that normal-API-sourced information, but I guess it's good if all the metrics from this process share a common prefix.

peterbourgon:
I'll promote https://github.com/peterbourgon/fastly-exporter/releases/tag/v7.0.0-beta.1 to v7.0.0 in a day or two, please give it a test and report any issues! /cc @mrnetops @SuperQ

Issue #84, opened 2021-12-04
iterating http_sd support towards generic discovery


mrnetops:
I believe we can make http_sd work generically out of the box without `relabel_configs`.

by explicitly setting the `__params_target` label per target (and ideally using the host header and port from the request to set the target:port) we can get service discovery to work without any jiggery pokery.

I used https://github.com/pagarme/static-response-server to host the following 
```
[
  {
    "targets": [
      "fastly-exporter:8080"
    ],
    "labels": {
      "__param_target": "0AizkuJPvMmqhulU7fXXXX"
    }
  },
  {
    "targets": [
      "fastly-exporter:8080"
    ],
    "labels": {
      "__param_target": "0KO5PPKDAMlzAQ22fsXXXX"
    }
  }
]
```

along with the following minimal `http_sd_configs`
```
  - job_name: 'fastly-exporter'

    http_sd_configs:
            - url: http://static-content:7070
``` 
and everything worked out of the box.

peterbourgon:
> by explicitly setting the __params_target label per target (and ideally using the host header and port from the request to set the target:port) we can get service discovery to work without

How should the fastly-exporter change?

peterbourgon:
> I used https://github.com/pagarme/static-response-server to host the following . . .

It should not be necessary to run a separate process in order to use the fastly-exporter with http_sd_config.

mrnetops:
> > I used https://github.com/pagarme/static-response-server to host the following . . .
> 
> It should not be necessary to run a separate process in order to use the fastly-exporter with http_sd_config.

Oh, that was literally just to test that the new http_sd endpoint json worked the way I wanted without having to diddle the golang code first.

I had already gotten your `relabel_configs` variant working first before looking to see how we could iterate it.

peterbourgon:
Oh! OK, sorry for presuming :)

peterbourgon:
So instead of `/sd` returning

```json
[
	{
		"targets": [
			"abc",
			"def"
		]
	}
]
```

you want

```json
[
	{
		"targets": [
			"localhost:8080"
		],
		"labels": {
			"__param_target": "abc"
		}
	},
	{
		"targets": [
			"localhost:8080"
		],
		"labels": {
			"__param_target": "def"
		}
	}
]
```

Is that right? If it works, this avoids some configuration in the prometheus.yaml — does it do anything beyond that?

[...]

PR #85, opened 2021-12-04
Simplify generic HTTP SD

peterbourgon:
This seems to work? Addresses #84, maybe? /cc @SuperQ @mrnetops

### DIFF
--- a/pkg/prom/registry.go
+++ b/pkg/prom/registry.go
@@ -130,12 +130,18 @@ func (r *Registry) handleIndex(w http.ResponseWriter, req *http.Request) {
 }
 
 func (r *Registry) handleServiceDiscovery(w http.ResponseWriter, req *http.Request) {
-	targets := r.serviceIDs()
+	type targetsLabels struct {
+		Targets []string          `json:"targets"`
+		Labels  map[string]string `json:"labels"`
+	}
 
-	response := []struct {
-		Targets []string `json:"targets"`
-	}{
-		{Targets: targets},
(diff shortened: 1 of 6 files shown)


pkg/prom/registry.go
```diff
@@ -130,12 +130,18 @@ func (r *Registry) handleIndex(w http.ResponseWriter, req *http.Request) {
...
+
+	var response []targetsLabels
 
-	response := []struct {
-		Targets []string `json:"targets"`
-	}{
-		{Targets: targets},
+	for _, serviceID := range r.serviceIDs() {
+		response = append(response, targetsLabels{
+			Targets: []string{serviceID},
```
peterbourgon, line 142: This needs to be present...

pkg/prom/registry.go
```diff
@@ -130,12 +130,18 @@ func (r *Registry) handleIndex(w http.ResponseWriter, req *http.Request) {
...
+			Labels:  map[string]string{"__param_target": serviceID},
```
peterbourgon, line 143: ...and this needs to be unique. Apparently. 

Tested with Prometheus 2.30.0.

peterbourgon:
Oops, nope, it doesn't actually work. Prometheus doesn't scrape the data properly.

peterbourgon:
Yeah, the thing is that source_label `__address__` is only knowable by Prometheus itself. You can't avoid that relabel config. So I'm not sure this juice is worth the squeeze.

PR #88, opened 2022-01-06
Refactor service_cache to use pagination

phamann:
### TL;DR
Refactors the service cache to iterate over the remote service list via pagination.

### Why?
As of early-2021 Fastly's [`/service`](https://developer.fastly.com/reference/api/services/service/#list-services) API moved to a paginated interface via query params instead of always returning the full list of services, the list now defaults to `20`. 

To make matters more confusing, the new default of `20` will only be returned for newer customer accounts created after `2021-02-01`, older customers will still get a full list if they request `/service` without any query params - which is likely why no users of this exporter have complained yet. 

This should also have a slight performance boost for the exporter (and the Fastly API :wink:), as some larger Fastly customers can have 1000s of services which result in very large HTTP/JSON responses.

### Note:
A lot of the diff is whitespace and thus better viewed with it disabled via: [`?w=1`](https://github.com/peterbourgon/fastly-exporter/pull/88/files?w=1)


peterbourgon:
Cool! Overall approach looks good. I'll propose a few changes, probably as a PR to this PR if that's OK?

phamann:
> Cool! Overall approach looks good. I'll propose a few changes, probably as a PR to this PR if that's OK?

@peterbourgon Hai 👋🏻  - Yes, that's absolutely fine and whatever is easiest for you!

peterbourgon:
Thank you for tolerating my pathologies.

Issue #90, opened 2022-01-08
Exporter should set it's own custom user-agent


phamann:
It would be nice if the exporter can set a custom user-agent string so that it can be identified by downstream APIs. Such as `User-Agent: fastly-exporter v1.2.3`.

I'm happy to take this work on next week, but detailing here before I forget.

peterbourgon:
We're already doing this [when querying rt.fastly.com](https://github.com/peterbourgon/fastly-exporter/blob/1e1277d4e643760037d8d6eea83afbb9e573b642/cmd/fastly-exporter/main.go#L289) so you can just do a similar thing for the API calls I guess.

peterbourgon:
Could we bake the user agent into the client itself?

peterbourgon:
@phamann @drhoden-fastly Take a look at `main` when you get a chance.

phamann:
Woop, perfect! Thank you, @peterbourgon for all your help getting this issue across the line. 

Are we able to cut a new release with the recent changes from #92 and #88?

peterbourgon:
https://github.com/peterbourgon/fastly-exporter/releases/v7.1.0 🖖

PR #91, opened 2022-01-10
Allow custom User Agent for all API invocations

drhoden-fastly:
### TL;DR

This PR implements the custom user agent request initiated by @phamann from https://github.com/peterbourgon/fastly-exporter/issues/90

### Why?

`http.RoundTripper` was selected to drive this solution because it was very obvious, configurable and is similar to the approach taken with `rt.WithUserAgent` injecting the user agent at runtime. 

@peterbourgon  suggested baking the user agent at build time. This was not explored as it is unclear if this meant adding to the linker Flag invocation in the Makefile or something else. Guidance is appreciated and I am happy to explore this solution or any mods to this PR (reference: https://github.com/peterbourgon/fastly-exporter/issues/90#issuecomment-1008150330).

👋🏿 @peterbourgon hope you are well and thanks for the review :) 

### Note:

n/a

### DIFF
--- a/cmd/fastly-exporter/main.go
+++ b/cmd/fastly-exporter/main.go
@@ -55,7 +55,7 @@ func main() {
 		fs.StringVar(&namespace, "namespace", "fastly", "Prometheus namespace")
 		fs.StringVar(&subsystem, "subsystem", "rt", "Prometheus subsystem")
 		fs.StringVar(&serviceShard, "service-shard", "", "if set, only include services whose hashed IDs modulo m equal n-1 (format 'n/m')")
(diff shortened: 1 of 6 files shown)


peterbourgon:
Thanks! I think this feature request moves the api.ServiceCache, api.DatacenterCache, rt.Manager, and rt.Subscriber closer together in certain respects, and reveals that they should probably be refactored a little bit. Let me do that real fast and I'll cc: you for a review.

drhoden-fastly:
neat thanks @peterbourgon. If I can clean up anything please lmk.

drhoden-fastly:
I am going to be stepping away for a bit, but I chatted with @phamann directly and he agreed to address any feedback in my absence. Thanks again for your work!

PR #94, opened 2022-03-04
Fix pagination link extraction

### LINKED ISSUE #93: Pagination is not working properly (not fetching every pages)
Hi !
After updating to latest version we noticed that we were missing some services, 
On version 7:
`level=debug component=api.fastly.com refresh_took=749.283548ms total_service_count=127 accepted_service_count=0
`
On version 6 or before:
`level=debug component=api.fastly.com refresh_took=2.052766648s total_service_count=7528 
`

We got 76 pages of services and the 127 services found by version 7 comes from only the first page and the last one, so the pagination in exporter is not working properly and not fetching each pages but only the first one and the last one.

Thank you !

phamann:
Fixes: #93 

Diff is mainly whitespace change, best best viewed with [`?w=1`](https://github.com/peterbourgon/fastly-exporter/pull/94/files?w=1)

### DIFF
--- a/pkg/api/link.go
+++ b/pkg/api/link.go
@@ -34,38 +34,40 @@ func uriFromLinks(links []string, k, v string) (string, bool) {
 
 func uriFromLink(link, k, v string) (string, bool) {
 	var rawuri string
-	for _, param := range strings.Split(link, ";") {
-		param = strings.TrimSpace(param)
-		if param == "" {
-			continue
-		}
+	for _, link := range strings.Split(link, ",") {
+		for _, param := range strings.Split(link, ";") {
+			param = strings.TrimSpace(param)
+			if param == "" {
+				continue
+			}
 
-		if rawuri == "" && param[0] == '<' && param[len(param)-1] == '>' {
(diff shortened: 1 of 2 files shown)


peterbourgon:
Aw beans okay. I'll add a test if it's not too hard, and will merge and cut a new release in a day or so.

phamann:
Woops, apologies @peterbourgon I have now included a test which is a single link header value containing multiple links separated by a `,` instead of multiple link header values as the other tests assumed.

pkg/api/link.go
```diff
@@ -34,38 +34,40 @@ func uriFromLinks(links []string, k, v string) (string, bool) {
...
-			continue
-		}
+	for _, link := range strings.Split(link, ",") {
+		for _, param := range strings.Split(link, ";") {
+			param = strings.TrimSpace(param)
+			if param == "" {
+				continue
+			}
 
-		if rawuri == "" && param[0] == '<' && param[len(param)-1] == '>' {
```
peterbourgon, line 43: 😬

peterbourgon (approved):

PR #96, opened 2022-03-24
Increase the service cache request page size from 100 to 1000

phamann:
### TL;DR
We have recently increased the maximum result page size of the `/service` endpoint from 100 to 1000. Therefore, this PR bumps the size the exporter uses to fetch the serivce list to `1000` and should hopefully reduce requests to the Fastly API by 10x.

### DIFF
--- a/pkg/api/service_cache.go
+++ b/pkg/api/service_cache.go
@@ -15,6 +15,10 @@ import (
 	"github.com/peterbourgon/fastly-exporter/pkg/filter"
 )
 
+// maxServicePageSize is the maximum amount of results that can be requested
+// from the api.fastly.com/service endpoint.
+const maxServicePageSize = 1000
+
 // Service metadata associated with a single service.
 // Also serves as a DTO for api.fastly.com/service.
 type Service struct {
@@ -92,7 +96,7 @@ func (c *ServiceCache) Refresh(ctx context.Context) error {
 	begin := time.Now()
 
 	var (
-		uri     = "https://api.fastly.com/service?page=1&per_page=100"
+		uri     = fmt.Sprintf("https://api.fastly.com/service?page=1&per_page=%d", maxServicePageSize)
 		total   = 0
 		nextgen = map[string]Service{}
 	)


pkg/api/service_cache.go
```diff
@@ -92,7 +96,7 @@ func (c *ServiceCache) Refresh(ctx context.Context) error {
 	begin := time.Now()
 
 	var (
-		uri     = "https://api.fastly.com/service?page=1&per_page=100"
+		uri     = fmt.Sprintf("https://api.fastly.com/service?page=1&per_page=%d", maxServicePageSize)
```
peterbourgon, line 99: Why not just hard-code it like before?

pkg/api/service_cache.go
```diff
@@ -15,6 +15,10 @@ import (
 	"github.com/peterbourgon/fastly-exporter/pkg/filter"
 )
 
+// maxServicePageSize is the maximum amount of results that can be requested
+// from the api.fastly.com/service endpoint.
+const maxServicePageSize = 1000
```
peterbourgon, line 20: Is this number defined or documented anywhere? If you reduce it, will this code start breaking?

pkg/api/service_cache.go
phamann: Just personal style preference, I like the explicitness and visibility for other readers of a constant for values like this. Very happy to have inline if you wish?

pkg/api/service_cache.go
phamann: It **_should_** be documented but it seems like it isn't. I can make sure that's fixed 😄

phamann: With regards to breakage, it looks like the internal pagination code defaults to using the current maximum instead of returning a 400 or similar if a request provides an out of bounds value. So the exporter will not break if this number gets reduced internally (which I doubt it will).

pkg/api/service_cache.go
peterbourgon: That's fine! Just wondering if there was another reason.

pkg/api/service_cache.go
peterbourgon: > It **_should_** be documented but it seems like it isn't. I can make sure that's fixed 😄

Sure! Gimme a ping (or PR) when it is, I'd like to update the doc comment here to point to that.

PR #98, opened 2022-05-10
cmd/fieldgen: use go:generate to generate pkg/gen/gen.go

shawnps:
Looking at https://github.com/fastly/fastly-exporter/pull/95 maybe the `gen` package is no longer needed though.


hermanschaaf (approved):
LGTM - I ran `go generate pkg/gen/doc.go` locally and confirmed that there are no differences with the generated code committed to the main branch. 

Perhaps we should document the process for regenerating the code, and/or add a Makefile entry?

PR #99, opened 2022-05-10
pkg/prom: add TestRegistryEndpoints cases for "text/html" and "application/json" accept headers

shawnps:


pkg/prom/registry_test.go
```diff
@@ -132,11 +146,23 @@ func TestRegistryEndpoints(t *testing.T) {
...
+		body := get(testRequest{path: "/", accept: "text/html"})
+		expect(strings.Contains(body, "AAA"), "AAA missing")
+		expect(strings.Contains(body, "BBB"), "BBB missing")
+	})
+
+	t.Run("index; accept:application/json", func(t *testing.T) {
+		body := get(testRequest{path: "/", accept: "application/json"})
+		expect(strings.Contains(body, "AAA"), "AAA missing")
+		expect(strings.Contains(body, "BBB"), "BBB missing")
+	})
```
hermanschaaf, line 170: Would it be worth asserting that the `application/json` case returns a response in JSON format?

shawnps: @hermanschaaf good idea 👍 I pushed an update, please take another look

hermanschaaf (approved):

PR #101, opened 2022-05-12
origin inspector

shawnps:
Rebased https://github.com/fastly/fastly-exporter/pull/95 onto main


cmd/fastly-exporter/main.go
```diff
@@ -106,6 +102,16 @@ func main() {
 		}
 	}
 
+	switch deprecatedSubsystem {
+	case "rt":
+		// good
+	case "origin":
+		level.Error(logger).Log("err", "-subsystem cannot be 'origin'")
```
hermanschaaf, line 109: In this case, it might be useful to tell the user what flag they should be using for Origin Inspector?

pkg/origin/metrics.go
```diff
@@ -0,0 +1,62 @@
...
+type Metrics struct {
+	RespBodyBytesTotal   *prometheus.CounterVec
+	RespHeaderBytesTotal *prometheus.CounterVec
+	ResponsesTotal       *prometheus.CounterVec
+	StatusCodeTotal      *prometheus.CounterVec
+	StatusGroupTotal     *prometheus.CounterVec
+	LatencySeconds       *prometheus.HistogramVec
+}
+
+// NewMetrics returns a new set of metrics registered to the registerer.
```
hermanschaaf: Nit: `registerer` 😅  I guess it fits with the -er interface pattern in Go but might I suggest using `registrar`?

pkg/origin/process.go
```diff
@@ -0,0 +1,62 @@
...
+				m.StatusGroupTotal.WithLabelValues(serviceID, serviceName, datacenter, origin, "4xx").Add(float64(stats.Status4xx))
+				m.StatusCodeTotal.WithLabelValues(serviceID, serviceName, datacenter, origin, "500").Add(float64(stats.Status500))
+				m.StatusCodeTotal.WithLabelValues(serviceID, serviceName, datacenter, origin, "501").Add(float64(stats.Status501))
+				m.StatusCodeTotal.WithLabelValues(serviceID, serviceName, datacenter, origin, "502").Add(float64(stats.Status502))
+				m.StatusCodeTotal.WithLabelValues(serviceID, serviceName, datacenter, origin, "503").Add(float64(stats.Status503))
+				m.StatusCodeTotal.WithLabelValues(serviceID, serviceName, datacenter, origin, "504").Add(float64(stats.Status504))
+				m.StatusCodeTotal.WithLabelValues(serviceID, serviceName, datacenter, origin, "505").Add(float64(stats.Status505))
+				m.StatusGroupTotal.WithLabelValues(serviceID, serviceName, datacenter, origin, "5xx").Add(float64(stats.Status5xx))
+
+				// Latency stats are clearly from xxx_bucket{le="v"} metrics,
```
hermanschaaf, line 33: This seems like a fair compromise to make the stats fit with Prometheus histogram metrics.

pkg/origin/response.go
```diff
@@ -0,0 +1,61 @@
...
+	Latency50to100      int `json:"latency_50_to_100ms"`      // Number of responses from origin with latency between 50 and 100 milliseconds.
+	Latency100to250     int `json:"latency_100_to_250ms"`     // Number of responses from origin with latency between 100 and 250 milliseconds.
+	Latency250to500     int `json:"latency_250_to_500ms"`     // Number of responses from origin with latency between 250 and 500 milliseconds.
+	Latency500to1000    int `json:"latency_500_to_1000ms"`    // Number of responses from origin with latency between 500 and 1,000 milliseconds.
+	Latency1000to5000   int `json:"latency_1000_to_5000ms"`   // Number of responses from origin with latency between 1,000 and 5,000 milliseconds.
+	Latency5000to10000  int `json:"latency_5000_to_10000ms"`  // Number of responses from origin with latency between 5,000 and 10,000 milliseconds.
+	Latency10000to60000 int `json:"latency_10000_to_60000ms"` // Number of responses from origin with latency between 10,000 and 60,000 milliseconds.
+	Latency60000plus    int `json:"latency_60000ms"`          // Number of responses from origin with latency of 60,000 milliseconds and above.
```
hermanschaaf: Aside: I hope that in the future we will have an endpoint to list all available metrics, which we would then be able to use here instead of hard-coding structs.

hermanschaaf: I notice that these are all `int`, whereas the fields in realtime/response.go use `uint64`. It doesn't matter too much, but for consistency we can also have the fields here use `uint64`.

hermanschaaf (approved):
LGTM, just left a couple of minor comments with thoughts as I went through this.

cmd/fastly-exporter/main.go
shawnps: @hermanschaaf I'm actually not sure if we need this. 

Maybe I'm misunderstanding the original intent but it seems like `-subsystem` was meant to be deprecated entirely (with `rt` accepted in the interim). In that case couldn't we just have anything but `-subsystem rt` error through the default clause below?

pkg/origin/process.go
```diff
@@ -0,0 +1,62 @@
...
+					0.500:  stats.Latency500to1000,
+					0.250:  stats.Latency250to500,
+					0.100:  stats.Latency100to250,
+					0.050:  stats.Latency50to100,
+					0.010:  stats.Latency10to50,
+					0.005:  stats.Latency5to10,
+					0.001:  stats.Latency1to5,
+					0.0005: stats.Latency0to1, // yolo
+				} {
+					for i := 0; uint64(i) < n; i++ {
```
hermanschaaf:
```suggestion
					for i := uint64(0); i < n; i++ {
```

This would declare `i` as `uint64`, rather than casting as part of the comparison

cmd/fastly-exporter/main.go
hermanschaaf: Yeah, I'm also not sure about the original intent but what you're suggesting sounds good to me too.

shawnps: @hermanschaaf looking at it again I think I get it -- the "origin" case calls `os.Exit(1)` but the default just logs a warning. So it's being explicit that "origin" isn't allowed. I'll keep it the way it is 👍

peterbourgon: `-subsystem` was originally user-specifiable, but with the addition of Origin Inspector, the intent is that the subsystem dimension of metric names has been "claimed" by the exporter. That is, all metrics captured and exported prior to this Origin Inspector PR are in the `rt` subsystem, and all Origin Inspector metrics are in the `origin` subsystem.

pkg/origin/metrics.go
peterbourgon: The comment reflects the type, which is `prometheus.Registerer` – don't shoot the messenger :) and please revert the change.

pkg/origin/response.go
peterbourgon: I wish it were possible! Alas, the mapping of the Origin Inspector metrics to viable Prometheus metrics isn't well-defined. In fact the _raison d'être_ of the fastly-exporter is to do exactly the sort of rote translation on display here, which (in Go) more or less demands concrete types like these.

PR #103, opened 2022-05-16
Adding edge/shield-specific hit and miss fields

mattcallahan:
Adds the following edge/shield-specific hit and miss fields:
- `edge_hit_requests`
- `edge_hit_resp_body_bytes`
- `edge_hit_resp_header_bytes`
- `edge_miss_requests`
- `edge_miss_resp_body_bytes`
- `edge_miss_resp_header_bytes`
- `shield_hit_requests`
- `shield_hit_resp_body_bytes`
- `shield_hit_resp_header_bytes`
- `shield_miss_requests`
- `shield_miss_resp_body_bytes`
- `shield_miss_resp_header_bytes`


shawnps:
Hi @mattcallahan thanks for the PR. I've reached out to the repo admins to see if we can enable CI actions for PRs. Will update this ticket when I hear back.

SuperQ:
One thing we typically do in Prometheus is to pick one of two, hits or misses, if there is also an "all" option.

So we don't actually need to exposes misses here, because the end user can derive misses from `edge_request - edge_hit_requests`.

It's also still possible to calculate hit/miss ratios with math like this:

```
1 - (edge_hit_requests / edge_request)
```

cmd/fieldgen/exporter_metrics.json
```diff
@@ -129,6 +135,12 @@
...
+    {"field_name": "ShieldHitRespHeaderBytesTotal",        "type": "Counter",   "metric_name": "shield_hit_resp_header_bytes_total",        "extra_labels": [],               "help": "Header bytes delivered for shield hits."},
+    {"field_name": "ShieldHitsTotal",                      "type": "Counter",   "metric_name": "shield_hit_requests",                       "extra_labels": [],               "help": "Number of requests that resulted in a hit at a shield."},
+    {"field_name": "ShieldMissRespBodyBytesTotal",         "type": "Counter",   "metric_name": "shield_miss_resp_body_bytes_total",         "extra_labels": [],               "help": "Body bytes delivered for shield misses."},
+    {"field_name": "ShieldMissRespHeaderBytesTotal",       "type": "Counter",   "metric_name": "shield_miss_resp_header_bytes_total",       "extra_labels": [],               "help": "Header bytes delivered for shield misses."},
+    {"field_name": "ShieldMissesTotal",                    "type": "Counter",   "metric_name": "shield_miss_requests",                      "extra_labels": [],               "help": "Number of requests that resulted in a miss at a shield."},
```
SuperQ, line 143: Counter metric names should end in `_total` by convention.

hermanschaaf:
@SuperQ

> One thing we typically do in Prometheus is to pick one of two, hits or misses, if there is also an "all" option.

This is likely to be good enough in many cases, but be aware that `edge_requests` is not necessarily equal to `edge_hit_requests + edge_miss_requests`--if I'm not mistaken the total `edge_requests` also includes passes, errors, pipe and synthetic responses, which are not counted as either `hit` or `miss`.

SuperQ:
@hermanschaaf Interesting, if there's more than hit and miss, it makes sense to have them as labels.

Really, it would be preferable if the upstream API provided something like this:

```
edge_requests{response="hit"} 1
edge_requests{response="miss"} 1
edge_requests{response="pass"} 1
...
```

This way `sum(edge_requests)` would include all valid responses.

leklund:
These fields were finally added in #131 and released in [v7.5.0](https://github.com/fastly/fastly-exporter/releases/tag/v7.5.0)

PR #105, opened 2022-05-23
cmd/fastly-exporter: move roundTripperFunc, userAgentTransport into main.go, allowing "go run cmd/fastly-exporter/main.go"

shawnps:


hermanschaaf (approved):

peterbourgon:
`go run` is meant as a convenience for trivial, generally single-file, script-ish Go programs. It's not really something that should be used with projects like this. It calculates the source that goes into the compilation unit in a way that's different than all other `go` tool subcommands, which is confusing -- hence this PR! :wink: And it doesn't take advantage of intermediate build caching to the extent that `go build` and/or `go install` do.

Anyone working on this project should be building-and-running via

```
go build ./cmd/fastly-exporter 
./fastly-exporter ...
```

or 

```
go install ./cmd/fastly-exporter
fastly-exporter ...
```

But if you _insist_ on using `go run` :wink: then you should be able to `go run cmd/fastly-exporter` — i.e. eliding the `main.go` — which should work without requiring the change in this PR?

PR #117, opened 2022-11-15
Add a random, exponential increasing delay when there are no results for a service

leklund:
The first delay will be between 2 and 11 seconds and it will max out
beween 130 and 175 seconds:

delay count: range of possible delay in seconds

```
1 => [2, 11]
2 => [10, 28]
3 => [30, 57]
4 => [68, 104]
5 (and up) => [130, 175]
```

### DIFF
--- a/pkg/rt/subscriber.go
+++ b/pkg/rt/subscriber.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/rand"
 	"net/http"
 	"net/url"
 	"strconv"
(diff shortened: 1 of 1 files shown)


pkg/rt/subscriber.go
```diff
@@ -40,6 +41,7 @@ type Subscriber struct {
 	metrics     *prom.Metrics
 	postprocess func()
 	logger      log.Logger
+	delayCount  int
```
crivera-fastly: `RunRealtime()` and `RunOrigins()` are both called on the same subscriber instance. I think we'll need separate delay counts for origins and billing.

leklund: ah. good point. fixed in [d4bc0a6](https://github.com/fastly/fastly-exporter/pull/117/commits/d4bc0a62bbf2971157f5dbd27efcbcbc3a34cd1e)

crivera-fastly (approved):

PR #118, opened 2022-11-16
Add product entitlement check for Origin Inspector before enabling the subscriber for origins.

leklund:
If origin inspector has not been enabled all requests for real time origin metrics will no data. This PR checks for the product before calling `subscriber.RunOrigins`.

It will not update once the exporter is running so when origin inspector has been purchased or enabled at the customer level, the exporter will need to be restarted.

### DIFF
--- a/cmd/fastly-exporter/main.go
+++ b/cmd/fastly-exporter/main.go
@@ -39,6 +39,7 @@ func main() {
 		metricAllowlist     stringslice
 		metricBlocklist     stringslice
 		datacenterRefresh   time.Duration
+		productRefresh      time.Duration
 		serviceRefresh      time.Duration
 		apiTimeout          time.Duration
 		rtTimeout           time.Duration
@@ -60,6 +61,7 @@ func main() {
 		fs.Var(&metricAllowlist, "metric-allowlist", "if set, only export metrics whose names match this regex (repeatable)")
 		fs.Var(&metricBlocklist, "metric-blocklist", "if set, don't export metrics whose names match this regex (repeatable)")
 		fs.DurationVar(&datacenterRefresh, "datacenter-refresh", 10*time.Minute, "how often to poll api.fastly.com for updated datacenter metadata (10m–1h)")
+		fs.DurationVar(&productRefresh, "product-refresh", 10*time.Minute, "how often to poll api.fastly.com for updated product metadata (10m–1h)")
 		fs.DurationVar(&serviceRefresh, "service-refresh", 1*time.Minute, "how often to poll api.fastly.com for updated service metadata (15s–10m)")
 		fs.DurationVar(&serviceRefresh, "api-refresh", 1*time.Minute, "DEPRECATED -- use service-refresh instead")
 		fs.DurationVar(&apiTimeout, "api-timeout", 15*time.Second, "HTTP client timeout for api.fastly.com requests (5–60s)")
@@ -126,6 +128,14 @@ func main() {
 		if datacenterRefresh > 1*time.Hour {
 			level.Warn(logger).Log("msg", "-datacenter-refresh cannot be longer than 1h; setting it to 1h")
 			datacenterRefresh = 1 * time.Hour
+		}
+		if productRefresh < 10*time.Minute {
+			level.Warn(logger).Log("msg", "-product-refresh cannot be shorter than 10m; setting it to 10m")
+			productRefresh = 10 * time.Minute
+		}
+		if productRefresh > 1*time.Hour {
+			level.Warn(logger).Log("msg", "-product-refresh cannot be longer than 1h; setting it to 1h")
+			productRefresh = 1 * time.Hour
 		}
 		if serviceRefresh < 15*time.Second {
 			level.Warn(logger).Log("msg", "-service-refresh cannot be shorter than 15s; setting it to 15s")
@@ -282,7 +292,7 @@ func main() {
 		})
 		g.Go(func() error {
 			if err := productCache.Refresh(context.Background()); err != nil {
-				level.Warn(logger).Log("during", "initial fetch of products", "err", err, "msg", "products API unavailable")
+				level.Warn(logger).Log("during", "initial fetch of products", "err", err, "msg", "products API unavailable, will retry")
 			}
 			return nil
 		})
@@ -333,6 +343,29 @@ func main() {
 				case <-ticker.C:
 					if err := datacenterCache.Refresh(ctx); err != nil {
 						level.Warn(apiLogger).Log("during", "datacenter refresh", "err", err, "msg", "the datacenter info metrics may be stale")
+					}
+				case <-ctx.Done():
+					return ctx.Err()
+				}
+			}
+		}, func(error) {
+			ticker.Stop()
+			cancel()
+		})
+	}
+	{
+		// Every productRefresh, ask the api.ProductCache to refresh
+		// data from the product entitlement endpoint.
+		var (
+			ctx, cancel = context.WithCancel(context.Background())
+			ticker      = time.NewTicker(productRefresh)
+		)
+		g.Add(func() error {
+			for {
+				select {
+				case <-ticker.C:
+					if err := productCache.Refresh(ctx); err != nil {
+						level.Warn(apiLogger).Log("during", "product refresh", "err", err, "msg", "the product entitlement data may be stale")
 					}
 				case <-ctx.Done():
 					return ctx.Err()
--- a/cmd/fastly-exporter/main.go
+++ b/cmd/fastly-exporter/main.go
@@ -1,3 +1,4 @@
+// Package main is the entry point for the fastly-exporter.
 package main
 
 import (
--- a/cmd/fastly-exporter/main.go
+++ b/cmd/fastly-exporter/main.go
@@ -281,7 +281,7 @@ func main() {
 			return nil
 		})
 		g.Go(func() error {
-			if err := productCache.Fetch(context.Background()); err != nil {
+			if err := productCache.Refresh(context.Background()); err != nil {
 				level.Warn(logger).Log("during", "initial fetch of products", "err", err, "msg", "products API unavailable")
 			}
 			return nil
--- a/cmd/fastly-exporter/main.go
+++ b/cmd/fastly-exporter/main.go
@@ -315,7 +315,7 @@ func main() {
 				rt.WithMetadataProvider(serviceCache),
 			}
 		)
-		manager = rt.NewManager(serviceCache, rtClient, token, registry, subscriberOptions, productCache.Products(), rtLogger)
+		manager = rt.NewManager(serviceCache, rtClient, token, registry, subscriberOptions, productCache, rtLogger)
 		manager.Refresh() // populate initial subscribers, based on the initial cache refresh
 	}
 
--- a/cmd/fastly-exporter/main.go
+++ b/cmd/fastly-exporter/main.go
@@ -261,6 +261,11 @@ func main() {
 		datacenterCache = api.NewDatacenterCache(apiClient, token)
 	}
 
+	var productCache *api.ProductCache
+	{
+		productCache = api.NewProductCache(apiClient, token, apiLogger)
+	}
+
 	{
 		var g errgroup.Group
 		g.Go(func() error {
@@ -275,6 +280,13 @@ func main() {
 			}
 			return nil
 		})
+		g.Go(func() error {
+			if err := productCache.Fetch(context.Background()); err != nil {
+				level.Warn(logger).Log("during", "initial fetch of products", "err", err, "msg", "products API unavailable")
+			}
+			return nil
+		})
+
 		g.Wait()
 	}
 
@@ -303,7 +315,7 @@ func main() {
 				rt.WithMetadataProvider(serviceCache),
 			}
 		)
-		manager = rt.NewManager(serviceCache, rtClient, token, registry, subscriberOptions, rtLogger)
+		manager = rt.NewManager(serviceCache, rtClient, token, registry, subscriberOptions, productCache.Products(), rtLogger)
 		manager.Refresh() // populate initial subscribers, based on the initial cache refresh
 	}
 
--- a/pkg/api/product_cache.go
+++ b/pkg/api/product_cache.go
@@ -12,13 +12,16 @@ import (
 )
 
 const (
-	Default         = "default"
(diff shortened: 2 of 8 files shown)


cmd/fastly-exporter/main.go
```diff
@@ -275,6 +280,13 @@ func main() {
 			}
 			return nil
 		})
+		g.Go(func() error {
+			if err := productCache.Fetch(context.Background()); err != nil {
```
crivera-fastly: Should we rename Fetch to Refresh to be consistent with the other caches?

pkg/api/product_cache.go
```diff
@@ -0,0 +1,87 @@
...
+
+		var response ProductAccess
+
+		if err := json.NewDecoder(resp.Body).Decode(&response); err != nil {
+			return fmt.Errorf("error decoding API product response: %w", err)
+		}
+
+		level.Debug(p.logger).Log("product", response.Name, "hasAccess", response.HasAccess)
+
+		p.products[response.Name] = response.HasAccess
```
crivera-fastly: We don't ever need to list the full set of products from the api response, right? If not then we don't care about the products we don't have access to. Maybe we should switch `products` to be `map[string]struct{}`. If the product key is in the map then we have access. We can rename `Products()` to `HasAccess(p string) bool`. If we do this then we can pass a ProductCache pointer directly to the manager and the `runOrigins()` func inside the manager goes away since it can just call `HasAccess()` directly where it's currently calling `runOrigins()`.

cmd/fastly-exporter/main.go
leklund: Yeah, that makes sense. I started with Fetch since it was only going to run once but `Refresh` is better for consistency.

pkg/api/product_cache.go
```diff
@@ -0,0 +1,91 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+)
+
+// ProductAccess models the response from the Fastly Product Entitlement API
+type ProductAccess struct {
```
crivera-fastly: It looks like `Product` is only used to be embedded in `ProductAccess`. This is a nit, but I would merge `ProductAccess` and `Product` to just `Product`.

pkg/api/product_cache.go
```diff
@@ -0,0 +1,89 @@
...
+
+		if response.HasAccess {
+			p.products[response.Meta.Name] = struct{}{}
+		}
+
+	}
+	return nil
+}
+
+// Products returns the list of products
```
crivera-fastly: We need to update this comment.

crivera-fastly (approved):

peterbourgon:
If you call Refresh only at program start and never again, it's not really a cache, it's just a value, you'd be better served by just fetching the values and providing them as a slice of strings or whatever to the service. If it's a cache then you should Refresh it on some schedule, and in that case it needs synchronization with an e.g. mutex.

pkg/api/product_cache.go
```diff
@@ -75,15 +84,23 @@ func (p *ProductCache) Refresh(ctx context.Context) error {
...
 			p.products[response.Meta.Name] = struct{}{}
+			p.mtx.Unlock()
 		}
 
 	}
 	return nil
 }
 
-// Products returns the list of products
+// HassAcces takes a product as a string and returns a boolean
```
crivera-fastly: s/HassAcces/HasAccess/

pkg/api/product_cache.go
```diff
@@ -75,15 +84,23 @@ func (p *ProductCache) Refresh(ctx context.Context) error {
...
+// based on the reponse from the Product API.
```
crivera-fastly: s/reponse/response/

pkg/api/product_cache.go
```diff
@@ -0,0 +1,106 @@
...
+			return fmt.Errorf("error decoding API product response: %w", err)
+		}
+
+		level.Debug(p.logger).Log("product", response.Meta.Name, "hasAccess", response.HasAccess)
+
+		if response.HasAccess {
+			p.mtx.Lock()
+			p.products[response.Meta.Name] = struct{}{}
+			p.mtx.Unlock()
+		}
```
peterbourgon: If a product goes from having access to losing access, where is it deleted from p.products?

pkg/rt/manager.go
```diff
@@ -149,3 +175,10 @@ type interrupt struct {
 	cancel func()
 	done   <-chan error
 }
+
+func parseKey(key string) (string, string) {
+	ks := strings.Split(key, sep)
+	return ks[0], ks[1]
```
peterbourgon: strings.Split does not guarantee that it's return value will have cardinality 2, so you can't access indices 0 and 1 without a check.

pkg/rt/manager.go
```diff
@@ -68,32 +72,48 @@ func (m *Manager) Refresh() {
...
 
-	for _, id := range m.managedIDsWithLock() {
-		level.Info(m.logger).Log("service_id", id, "subscriber", "stop")
-		irq := m.managed[id]
-		irq.cancel()
-		err := <-irq.done
-		delete(m.managed, id)
-		level.Debug(m.logger).Log("service_id", id, "interrupt", err)
-	}
+		for _, key := range m.managedIDsWithLock() {
```
peterbourgon: If this function returns keys, it should be renamed to managedKeysWithLock.

pkg/rt/manager.go
```diff
@@ -68,32 +72,48 @@ func (m *Manager) Refresh() {
...
+			id, managedProduct := parseKey(key)
+			if managedProduct != product {
+				continue
+			}
+
+			level.Info(m.logger).Log("service_id", id, "type", product, "subscriber", "stop")
```
peterbourgon, line 97: The Log method of the logger interface requires an even number of parameters: always key/val pairs.

pkg/api/product_cache.go
```diff
@@ -0,0 +1,106 @@
...
+type ProductCache struct {
+	client HTTPClient
+	token  string
+	logger log.Logger
+
+	mtx      sync.Mutex
+	products map[string]struct{}
+}
+
+// NewProductCache returns an empty cache of Product information. Use the Fetch method
```
crivera-fastly: s/Fetch/Refresh/

pkg/rt/manager.go
```diff
@@ -68,32 +72,48 @@ func (m *Manager) Refresh() {
...
-			level.Debug(m.logger).Log("service_id", id, "subscriber", "maintain")
-			nextgen[id] = irq // move
-			delete(m.managed, id)
-		} else {
-			level.Info(m.logger).Log("service_id", id, "subscriber", "create")
-			nextgen[id] = m.spawn(id)
+	for _, product := range api.Products {
+		if m.productCache.HasAccess(product) {
+			for _, id := range m.ids.ServiceIDs() {
+				key := id + sep + product
```
crivera-fastly: I'm not sure this is worth it, but you can define a struct with these key components instead of shoving them into a string:

```
type ProductKey struct {
    ServiceID string
    Product string
}
```

Instances of this struct can be directly used as keys to the `managed` and `nextgen` maps while also giving you direct access to the components without splitting strings. This lets us get rid of the string concatenation and `parseKey()`.

leklund: Yeah, this is a really good idea. I wasn't happy with the separator and splitting the strings. I clean this up next week when I get back.

pkg/api/product_cache.go
leklund: Thanks for catching all my typos 🤦

pkg/api/product_cache.go
leklund: Right now it doesn't. We were more concerned about customers being able to export metrics when activating traffic without having to restart the exporter. I'll add delete to the Refresh.

pkg/rt/manager.go
```diff
@@ -102,47 +124,59 @@ func (m *Manager) Refresh() {
 
 // Active returns the set of service IDs currently being managed.
 // Mostly useful for tests.
-func (m *Manager) Active() (serviceIDs []string) {
+func (m *Manager) Active() []string {
+	serviceIDs := []string{}
```
crivera-fastly: This is a small thing, but we know what the size of `serviceIDs` is going to be. We should call `m.managedKeysWithLock()` and use the length of the return value as a parameter to `make()` to create `serviceIDs`.

pkg/rt/manager.go
```diff
@@ -67,33 +75,47 @@ func (m *Manager) Refresh() {
...
 
-	for _, id := range m.managedIDsWithLock() {
-		level.Info(m.logger).Log("service_id", id, "subscriber", "stop")
-		irq := m.managed[id]
-		irq.cancel()
-		err := <-irq.done
-		delete(m.managed, id)
-		level.Debug(m.logger).Log("service_id", id, "interrupt", err)
-	}
+		for _, key := range m.managedKeysWithLock() {
```
crivera-fastly: Should we iterate over `m.managed` directly here instead of using `m.managedKeysWithLock()`? We're already holding the lock. ` managedKeysWithLock()` gives us a slice sorted by service id, but we don't need a slice or the results to be sorted.  We just need to iterate over the subscriber keys that are still left in `m.managed`.

pkg/rt/manager.go
```diff
@@ -102,47 +124,59 @@ func (m *Manager) Refresh() {
...
+	}
+
 	return interrupt{cancel, done}
 }
 
-func (m *Manager) managedIDsWithLock() []string {
-	ids := make([]string, 0, len(m.managed))
-	for id := range m.managed {
-		ids = append(ids, id)
+func (m *Manager) managedKeysWithLock() []subscriberKey {
```
crivera-fastly: I wonder if we need this function all. The only users are in this file and none of them need a sorted slice for anything.

pkg/rt/manager.go
```diff
@@ -67,33 +79,47 @@ func (m *Manager) Refresh() {
...
 
-	for _, id := range m.managedIDsWithLock() {
-		level.Info(m.logger).Log("service_id", id, "subscriber", "stop")
-		irq := m.managed[id]
-		irq.cancel()
-		err := <-irq.done
-		delete(m.managed, id)
-		level.Debug(m.logger).Log("service_id", id, "interrupt", err)
-	}
+		for key := range m.managed {
```
crivera-fastly: If you change this line to `for key, irq := range m.managed` then you don't need the lookup on line 105.

crivera-fastly (approved):

pkg/api/product_cache.go
```diff
@@ -0,0 +1,108 @@
...
+	if product == Default {
+		return true
+	}
+	p.mtx.Lock()
+	defer p.mtx.Unlock()
+	if v, ok := p.products[product]; ok {
+		return v
+	}
+	return true
+}
```
peterbourgon, line 108: Will `p.HasAccess("abcxyz")` return true?

peterbourgon: More generally is access to a given product default-allow?

pkg/rt/manager.go
```diff
@@ -17,13 +17,23 @@ type ServiceIdentifier interface {
...
 // the method of the prom.Registry which yields a set of Prometheus metrics for
 // a specific service ID.
 type MetricsProvider interface {
 	MetricsFor(serviceID string) *prom.Metrics
 }
 
+type subscriberKey struct {
+	serviceID string
+	product   string
+}
```
peterbourgon, line 35: If product should be part of the subscriber key like this says, that would mean there would be different subscriber types for each product, e.g. a subscriberKey with serviceID=X and product=foo would represent an e.g. fooSubscriber, and a subscriberKey with serviceID=X and product=bar would map to an e.g. barSubscriber. But there's only a single subscriber type, it's parameterized on service ID, and it has methods for both (all) of the product features. That means "enabled products" is a property of that subscriber type, not the key. So this doesn't really work.

If you want distinct subscribers for unique service ID + product combinations, then product needs to be part of both the subscriberKey and a field in the subscriber struct, which is consumed by all of the methods on that struct that interact with the API. But if you don't put product in the subscriber struct, then you can't make product part of the subscriber key, and you have to manage products at a layer above individual subscribers.

pkg/api/product_cache.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+)
+
+const (
+	// Default is the standard real-time stats available to all services
+	Default = "default"
+	// OriginInspector is the product name used to determine access to Origin Inspector via the entitlement API
+	OriginInspector = "origin_inspector"
+)
```
peterbourgon, line 19: These identifiers resolve to `api.Default` and `api.OriginInspector` which are somewhat ambiguous.

```go
const (
	// ProductDefault represents the standard real-time stats available to all services.
	ProductDefault = "default"

	// ProductOriginInspector represents the origin inspector stats available via the
	// entitlement API.
	ProductOriginInspector = "origin_inspector"
)
```

pkg/api/product_cache.go
```diff
@@ -0,0 +1,108 @@
...
+// Products is the slice of available products supported by real-time stats.
+var Products = []string{Default, OriginInspector}
+
+// Product models the response from the Fastly Product Entitlement API.
+type Product struct {
+	HasAccess bool `json:"has_access"`
+	Meta      struct {
+		Name string `json:"id"`
+	} `json:"product"`
+}
```
peterbourgon, line 30:
```go
//  Product models the response from the Fastly Product Entitlement API.
 type Product struct {
 	HasAccess bool `json:"has_access"`
 	Product   struct {
 		ID string `json:"id"`
 	} `json:"product"`
 }
```

pkg/rt/manager.go
```diff
@@ -17,13 +17,23 @@ type ServiceIdentifier interface {
 	ServiceIDs() []string
 }
 
+// ProductCache represents the api.ProductCache behavior.
+type ProductCache interface {
+	HasAccess(string) bool
+}
```
peterbourgon, line 23:
```go
// HasAccesser models the read side of an api.ProductCache.
type HasAccesser interface {
	HasAccess(string) bool
}
```

pkg/api/product_cache.go
leklund: This much clearer, thank you.

PR #137, opened 2023-06-14
Fix the product-refresh setting check for the new default of 24h

froesef:
The new default setting is 24h but the productRefresh gets set to 24h if the value is larger than 1h.


leklund:
@froesef Thank you! I'll get a new release out shortly!

PR #153, opened 2023-11-11
Implement aggregated datacenter metrics

### LINKED ISSUE #152: Reduce output size of metrics endpoint
# Problem

Currently, when collecting stats for `201` services after running the exporter for `13` days with `12` shards the metrics endpoint output size is as follows:

| Shard     | Services | Payload (KB) |
| --------- | -------- | ------------ |
| 1         | 18       | 30,792       |
| 2         | 25       | 55,378       |
| 3         | 16       | 40,243       |
| 4         | 15       | 29,123       |
| 5         | 21       | 34,345       |
| 6         | 22       | 40,100       |
| 7         | 10       | 19,948       |
| 8         | 19       | 47,790       |
| 9         | 11       | 20,234       |
| 10        | 15       | 40,499       |
| 11        | 19       | 37,366       |
| 12        | 19       | 29,092       |
| **Total** | **210**  | **424,910**  |

With a scrape interval of `60` seconds the bandwidth requirement becomes `7,082` KB/s. In terms of storage requirements, this is `424,910` KB * `60` mins * `24` hours = `584` GB of raw data per day.

This can cause considerable impact on Prometheus scraping performance as this is a very large payload.

# Proposal

Currently, each datacenter is a label which multiplies the number of each metric. When combined with a metric that has a `status_code` label this can explode the number of metrics returned.

A possible solution to reduce the output size of the metrics endpoint would be to aggregate the datacenter. 

Analysis of how this might impact the output size for the earlier example is as follows:

| Shard     | Services | Payload (KB) |
| --------- | -------- | ------------ |
| 1         | 17       | 645          |
| 2         | 25       | 934          |
| 3         | 16       | 607          |
| 4         | 14       | 531          |
| 5         | 21       | 796          |
| 6         | 21       | 786          |
| 7         | 10       | 394          |
| 8         | 18       | 686          |
| 9         | 10       | 398          |
| 10        | 15       | 582          |
| 11        | 19       | 718          |
| 12        | 15       | 569          |
| **Total** | **201**  | **7,646**    |

With a scrape interval of `60` seconds the bandwidth requirement becomes `127` KB/s. In terms of storage requirements, this is `7,646` KB * `60` mins * `24` hours = `11` GB of raw data per day. 

A comparison to the results with having individual datacenter metrics shows the following improvements: 

| Datacenter | Payload (KB) | Rate (KB/s) | Storage (Daily in GB) | Reduction |
| ---------- | ------------ | ----------- | --------------------- | --------- |
| Individual | 424,910      | 7082        | 584                   |           |
| Aggregated | 7,646        | 127         | 11                    | **98**%   |

A side effect of having aggregated datacenter metrics would be the memory consumption should be reduced. It it hard to determine the exact impact but there should certainly be some improvements.

# Conclusion

Aggregated data center metrics would provide an option for users that wish to reduce the metrics endpoint output size. By providing this as an option (not the default) this would allow users to decide if the benefits of reducing the output size outweigh the loss of inidivual datacenter metrics.

mikelorant:
Add flag to use aggregated datacenter metrics instead of per-datacenter.

Closes: #152


mikelorant:
@leklund Would you be able to do a preliminary review, please?

leklund:
@mikelorant Sorry this has been sitting for so long. I'll take a look at this early next week.

README.md
```diff
@@ -119,6 +119,11 @@ Different flags (for the same filter target) combine with AND semantics. For
 example, `-metric-allowlist 'bytes_total$' -metric-blocklist imgopto` would only
 export metrics whose names ended in bytes_total, but didn't include imgopto.
 
+### Aggregate Datacenter
+
+Datacenter metrics can be aggregated to reduce the output size of the metrics
+endpoint by using the `-aggregate-datacenter` flag.
+
```
leklund: I think the name is slightly confusing. The intention seems to be to aggegrate without the datacenter dimension. I think we could more explicit as to what the flag is doing.

My suggestion would be to name the flag `-without-datacenter` or `-without-datacenter-dimension` for the extra verbosity

Then this part of the README might instead read:

```suggestion
### Removing the Datacenter Dimension

Metrics, by default, are aggregated by service and datacenter. The datacenter dimension can be removed to reduce the output size of the metrics
endpoint by using the `-without-datacenter-dimension` flag.

```

cmd/fastly-exporter/main.go
```diff
@@ -67,6 +69,7 @@ func main() {
 		fs.DurationVar(&serviceRefresh, "api-refresh", 1*time.Minute, "DEPRECATED -- use service-refresh instead")
 		fs.DurationVar(&apiTimeout, "api-timeout", 15*time.Second, "HTTP client timeout for api.fastly.com requests (5–60s)")
 		fs.DurationVar(&rtTimeout, "rt-timeout", 45*time.Second, "HTTP client timeout for rt.fastly.com requests (45–120s)")
+		fs.BoolVar(&aggregateDatacenter, "aggregate-datacenter", false, "Use aggregated data rather than per-datacenter")
```
leklund: The description does make the what the flag does a lot clearer but I still think it would be easier to understand if it was renamed.

pkg/rt/subscriber.go
```diff
@@ -223,7 +230,13 @@ func (s *Subscriber) queryRealtime(ctx context.Context, ts uint64) (currentName
 			s.rtDelayCount = 0
 			result = apiResultSuccess
 		}
-		realtime.Process(&response, s.serviceID, name, version, s.metrics.Realtime)
+
+		if s.aggregateDatacenter {
+			realtime.ProcessAggregated(&response, s.serviceID, name, version, s.metrics.Realtime)
+		} else {
+			realtime.Process(&response, s.serviceID, name, version, s.metrics.Realtime)
+		}
```
leklund: Instead of the conditional here maybe add an arg to Process?

```
realtime.Process(&response, s.serviceID, name, version, s.metrics.Realtime, withoutDatacenter)
```

pkg/realtime/process.go
```diff
@@ -10,227 +10,246 @@ import (
...
-			m.WebsocketConnTimeMsTotal.WithLabelValues(serviceID, serviceName, datacenter).Add(float64(stats.WebsocketConnTimeMs))
-			m.WebsocketReqBodyBytesTotal.WithLabelValues(serviceID, serviceName, datacenter).Add(float64(stats.WebsocketReqBodyBytes))
-			m.WebsocketReqHeaderBytesTotal.WithLabelValues(serviceID, serviceName, datacenter).Add(float64(stats.WebsocketReqHeaderBytes))
-			m.WebsocketRespBodyBytesTotal.WithLabelValues(serviceID, serviceName, datacenter).Add(float64(stats.WebsocketRespBodyBytes))
-			m.WebsocketRespHeaderBytesTotal.WithLabelValues(serviceID, serviceName, datacenter).Add(float64(stats.WebsocketRespHeaderBytes))
-			processHistogram(stats.MissHistogram, m.MissDurationSeconds.WithLabelValues(serviceID, serviceName, datacenter))
-			processObjectSizes(stats.ObjectSize1k, stats.ObjectSize10k, stats.ObjectSize100k, stats.ObjectSize1m, stats.ObjectSize10m, stats.ObjectSize100m, stats.ObjectSize1g, m.ObjectSizeBytes.WithLabelValues(serviceID, serviceName, datacenter))
+			processDatacenter(&datacenter, serviceID, serviceName, serviceVersion, stats, m)
```
leklund: If you add withoutDatacenter to the func signature then you might be able to simplify:

```
func Process(response *Response, serviceID, serviceName, serviceVersion string, m *Metrics, withoutDatacenter bool) {
	labels := prometheus.Labels{
		"service_id":   serviceID,
		"service_name": serviceName,
	}
	for _, d := range response.Data {
		if withoutDatacenter {
			process(labels, d.Aggregated, m)
		} else {
			for datacenter, stats := range d.Datacenter {
				labels["datacenter"] = datacenter
				process(labels, stats, m)
			}
		}
	}
}

func process(labels prometheus.Labels, stats Datacenter, m *Metrics) {
	 m.AttackBlockedReqBodyBytesTotal.With(mergeLabels(labels)).Add(float64(stats.AttackBlockedReqBodyBytes))
	 m.TLSTotal.With(mergeLabels(labels, prometheus.Labels{"tls_version": "1.0"})).Add(float64(stats.TLSv10))
	//etc
}
```

pkg/realtime/process.go
```diff
@@ -259,3 +278,17 @@ func processObjectSizes(n1k, n10k, n100k, n1m, n10m, n100m, n1g uint64, obs prom
...
+	for k, v := range labels {
+		merged[k] = v
+	}
+
+	for k, v := range extraLabels {
+		merged[k] = v
+	}
+
+	return merged
+}
```
leklund: If you make this variadic you won't have to pass a `nil` with every call

```suggestion
func mergeLabels(labels ...prometheus.Labels) prometheus.Labels {
	merged := make(map[string]string)

	for _, l := range labels {
		for k, v := range l {
			merged[k] = v
		}
	}
	return merged
}
```

pkg/realtime/metrics.go
```diff
@@ -200,197 +200,210 @@ type Metrics struct {
 	WebsocketRespHeaderBytesTotal          *prometheus.CounterVec
 }
 
+type DatacenterAggregate bool
+
+const (
+	AggregateDatacenter  DatacenterAggregate = true
+	IndividualDatacenter DatacenterAggregate = false
+)
```
leklund: Besides renaming these constants, they should probably live outside the realtime package so that they can be used to the origin/domain inspector metrics packages as well.

leklund: And it you take my naming suggestion then these might be named:
```
WithoutDatacenter(Label|Dimension)? 
WithDatacenter(Label|Dimension)?
```

pkg/prom/registry_test.go
```diff
@@ -22,7 +23,7 @@ func TestRegistryEndpoints(t *testing.T) {
 		namespace        = "fastly"
 		subsystem        = "rt"
 		metricNameFilter = filter.Filter{}
-		registry         = prom.NewRegistry(version, namespace, subsystem, metricNameFilter)
+		registry         = prom.NewRegistry(version, namespace, subsystem, metricNameFilter, realtime.IndividualDatacenter)
```
leklund: This might not be the right place but I think it would be good to have a test for the AggregatedDatacenter (WithoutDatacenter) option.

pkg/realtime/metrics.go
```diff
@@ -421,3 +434,13 @@ func getName(c prometheus.Collector) string {
...
+
+func addLabel(ls []string, l string) []string {
+	ns := make([]string, len(ls))
+
+	copy(ns, ls)
+
+	ns = append(ns, l)
+
+	return ns
+}
```
leklund: In the other subsystems, I *think* we have cases were we'll need to add more than one label to the label set. So this might be more reusable as a variadic func:

```suggestion
func addLabel(ls []string, l ...string) []string {
	ns := make([]string, len(ls), len(ls)+len(l))
	copy(ns, ls)

	return append(ns, l...)
}
```

leklund (commented):
@mikelorant -- apologies for letting this sit so long. Thank you for the submission! Just FYI I'll be out of the office for the next 10 days but I will check back in when I return.

I've left a few comments with some minor suggestions including a naming change. I'm not convinced that `-without-datacenter-dimension` is the best name for the flag. Since it's removing the label maybe it could be `-without-datacenter-label`? I'm open to suggestions!

This will also need to be added to `domain/(metrics|process).go` and `origin/(metrics|process).go` as well so that the flag affects all the subsystems. Once the implentation is complete this can be added to origins/domains.

mikelorant:
@leklund Thanks for the all the feedback!

All your points make sense and I am happy to make all the changes you requested. I'll expect to have your requested changes implemented by the time you get back.

mikelorant:
Breaking this pull request up into smaller pieces.

- [Refactor labels used in realtime package][1]

[1]: https://github.com/fastly/fastly-exporter/pull/167

pkg/realtime/metrics.go
mikelorant: > Besides renaming these constants, they should probably live outside the realtime package so that they can be used to the origin/domain inspector metrics packages as well.

Could you recommend a package to add these to? I've experimented with using the `prom` package but am not sure if that is the right place.

pkg/realtime/metrics.go
mikelorant: You will see this change applied in pull request #167. 👍

pkg/realtime/process.go
mikelorant: This change has been used in pull request #167. 👍

pkg/realtime/process.go
mikelorant: I have added the `process` function to pull request #167. 👍

mikelorant: My local branch now has the `Process` function work this way. Thanks for the tip.

pkg/rt/subscriber.go
mikelorant: Implemented in my local branch. 👍

README.md
mikelorant: I think the naming is especially difficult because it has already been termed `Aggregated`.
https://github.com/fastly/fastly-exporter/blob/f6a1c512da5ae7530d351fadcfd0b00e766f1c9f/pkg/realtime/response.go#L9-L13

The way I understand this is we two choices for metrics:

- Per Datacenter
- Aggregated

I have a slight problem with using `without` as a prefix because this means changing the default state would be very confusing. My plan is to pitch that the default should not include per datacenter metrics knowing now how much extra data this creates. We back ourselves into a corner with this name.

leklund: That makes sense re: naming. I think my confusion stems from using both `aggregate` _and_ `datacenter`. I read that as aggregating **by** datacenter. Maybe `aggregated-only`?

I'm reluctant to change the default behavior as that would be a pretty major breaking change and I know many users rely on the per-datacenter metrics.

leklund:
@mikelorant Thanks for the new PR! I spent some time looking at this today and had some thought/ideas. Left a summary at the top because it got long.

### Summary

I had a realization that instead of adding complexity to remove the datacenter label you could instead just use a constant for the datacenter and it will have the same desired effect of reducing the /metrics response size. It could also skip the datacenter iteration completely and just use the `Aggreagate` field from the Response struct (that's the last sample code at the very bottom).

### Details


I was curious about performance with all the extra map allocations. I wrote a simple benchmark for Process():

`main` branch:

```
pkg: [github.com/fastly/fastly-exporter/pkg/realtime](http://github.com/fastly/fastly-exporter/pkg/realtime)
BenchmarkProcess-12    	    2840	    419057 ns/op	    1883 B/op	      57 allocs/op
```

`mikelorant:refactor/labels` using mergeLabels:

```
pkg: [github.com/fastly/fastly-exporter/pkg/realtime](http://github.com/fastly/fastly-exporter/pkg/realtime)
BenchmarkProcess-12    	    1012	    999219 ns/op	  264528 B/op	    1640 allocs/op
```

I haven't tested out the real-world implications but it could be noticeable if you happen to use the exporter with a large number of service.

I did a quick pass to use a slices instead of the map and it's better:

```
pkg: [github.com/fastly/fastly-exporter/pkg/realtime](http://github.com/fastly/fastly-exporter/pkg/realtime)
BenchmarkProcess-12    	    2558	    465463 ns/op	   74551 B/op	     815 allocs/op
```

That last benchmark used this code:
```
func Process(response *Response, serviceID, serviceName, serviceVersion string, m *Metrics) {
	labels := make([]string, 3, 3)

	labels[0] = serviceID
	labels[1] = serviceName

	for _, d := range response.Data {
		for datacenter, stats := range d.Datacenter {
			labels[2] = datacenter
			process(labels, stats, m)
		}
	}
}

func process(labels []string, stats Datacenter, m *Metrics) {
	...
	m.StatusCodeTotal.WithLabelValues(append(labels, "200")...).Add(float64(stats.Status200))
	...
}
```


I also tried a version where I used the `prometheus.Labels` map and would add the extra labels and then delete the key as after use. It's not very elegant but was slightly better memory-wise than the slices.

```
pkg: [github.com/fastly/fastly-exporter/pkg/realtime](http://github.com/fastly/fastly-exporter/pkg/realtime)
BenchmarkProcess-12    	    1513	    787019 ns/op	   31202 B/op	     241 allocs/op
```

It looked like this:

```
labels["status_group"] = "1xx"
m.ComputeRespStatusTotal.With(labels).Add(float64(stats.ComputeRespStatus1xx))
... etc.
delete(labels, "status_group")
```

---

While exploring ways to optimize this I had a realization. This code is still iterating over the **per-datacenter** metrics and incrementing counters. Instead I think it could just use the `Aggregated` metrics that are returned by the real-time API and only increment once. The datacenter label could still exist just with the value of "aggregated". This means not adding complexity to the metric initialization and we can retain the same Process() func. In fact, the simplest approach would be to just change the datacenter name to "aggregated" and keep the label.


```
const aggdc = "aggregated"

func Process(response *Response, serviceID, serviceName, serviceVersion string, m *Metrics, aggOnly bool) {
	for _, d := range response.Data {
		for datacenter, stats := range d.Datacenter {
			if aggOnly {
				datacenter = aggdc
			}
		// increment all the metrics
	}
}
```

or using `Aggregate`: (**this would be my suggestion as we skip all the iteration**)

```
const aggdc = "aggregated"

func Process(response *Response, serviceID, serviceName, serviceVersion string, m *Metrics, aggOnly bool) {
 	 for _, d := range response.Data {
		if aggOnly {
			process(serviceID, serviceName, aggdc, stats, m)
		} else {
			for datacenter, stats := range d.Datacenter {
				process(serviceID, serviceName, datacenter, stats, m)
			}
		}
	}
}

func process(serviceID, serviceName, datacenter string, stats Datacenter, m *Metrics) {
	// increment all the things 
	 m.ComputeRespStatusTotal.WithLabelValues(serviceID, serviceName, datacenter, "1xx").Add(float64(stats.ComputeRespStatus1xx))
	// etc
}
```

leklund:
Since my last comment was very long and rambling I thought it might help to put my idea into code:

https://github.com/fastly/fastly-exporter/compare/agg-only?expand=1

mikelorant:
On first pass, I totally agree with you on the new approach. My initial implementation was about just setting the datacenter to `aggregated` but I was expecting to be rejected because it seemed lazy and instead decided to do the proper implementation of removing the label.

How much refinement do we need to your branch to turn it into something functional? I assume we need to make it work for domain and origin metrics?

I really am hoping to convince you based on the metrics that defaulting to per datacenter is a bad idea. The cardinality explosion is so major, that if you have hundreds of active services you are going to swap yourself in metrics. Add this in to the cost of using domain and origin inspector and you are multiplying out yet again.

This is the type of breaking change that I feel would be beneficial to all users. Make them opt-in to datacenter metrics if they need that detail. I doubt most realise how costly it is and likely doesn't provide tangible benefits.

leklund:
> On first pass, I totally agree with you on the new approach. My initial implementation was about just setting the datacenter to `aggregated` but I was expecting to be rejected because it seemed lazy and instead decided to do the proper implementation of removing the label.

While it does seem a little off to use "aggregated" as the datacenter I do like the simplicity of that implementation. I also like that it doesn't break existing dashboards that might rely on the datacenter label. I agree that the ideal case would probably be to remove the label completely but I try to be pragmatic about these things.

> How much refinement do we need to your branch to turn it into something functional? I assume we need to make it work for domain and origin metrics?

We need to update the Readme/documenation, add support for domains/origins, and add any test cases I missed. 

> I really am hoping to convince you based on the metrics that defaulting to per datacenter is a bad idea. The cardinality explosion is so major, that if you have hundreds of active services you are going to swap yourself in metrics. Add this in to the cost of using domain and origin inspector and you are multiplying out yet again.
> 
> This is the type of breaking change that I feel would be beneficial to all users. Make them opt-in to datacenter metrics if they need that detail. I doubt most realise how costly it is and likely doesn't provide tangible benefits.

It's definitely a high-cardinality set of metrics and I'm sure there are other users like yourself that would benefit from the smaller set of aggregate-only metrics. But I also know users that rely on the per datacenter information so I am reluctant to make this breaking change.

Even if this is not the default I think we can add a prominent note in the README on using the new flag to reduce cardinality.

mikelorant:
> We need to update the Readme/documenation, add support for domains/origins, and add any test cases I missed.

Sounds like this is the path forward then.

Happy to help with copying the implementation and adding the same code to both origin and domain.

leklund:
> Happy to help with copying the implementation and adding the same code to both origin and domain.

Thank you, that would be great! Feel free to copy/modify what I quickly wrote and add it origins/domains.

mikelorant:
Have created the initial pull requests to prepare for the feature:

- #168 
- #169 

I was then able to take your branch and extend it to add support for the domain and origin systems. I did some minor refactoring to reduce indentation and remove the global.

You can see the results on my [feature/aggregate-only][1] branch.

Be aware, this branch by itself does not pass unit tests until we merge #168. My plan is that once both #168 and #169 are merged, is to update this existing pull request with the code from that branch and **NOT** open a new pull request.

[1]: https://github.com/mikelorant/fastly-exporter/compare/refactor/process...mikelorant:fastly-exporter:feature/aggregate-only

mikelorant:
This is now the new implementation as we discussed.

Note that there is no documentation yet, if you are happy with this approach, I will add them.

mikelorant:
Outstanding tasks:

- [x] Add documentation.

mikelorant:
Preliminary documentation (modified version of what we previously had) was added. I think it would be wise to educate the users of the benefit of enabling this feature and how much of an impact this can have on **reducing** the number of time series emitted.

pkg/rt/subscriber.go
```diff
@@ -71,6 +72,11 @@ func WithPostprocess(f func()) SubscriberOption {
 	return func(s *Subscriber) { s.postprocess = f }
 }
 
+// TODO add doc comment
```
mikelorant: - [x] Update comment.

pkg/rt/subscriber.go
```diff
@@ -71,6 +72,13 @@ func WithPostprocess(f func()) SubscriberOption {
 	return func(s *Subscriber) { s.postprocess = f }
 }
 
+// WithAggregateOnly sets whether aggregate metrics are emitted instead of
```
mikelorant: @leklund Is this wording acceptable? Feel free to provide an update if you feel this can be explained better (I am not a wordsmith 😄).

README.md
```diff
@@ -119,6 +119,11 @@ Different flags (for the same filter target) combine with AND semantics. For
 example, `-metric-allowlist 'bytes_total$' -metric-blocklist imgopto` would only
 export metrics whose names ended in bytes_total, but didn't include imgopto.
 
+### Aggregate Datacenter
```
mikelorant: @leklund Recommendations for the type of content we want mentioned here?

README.md
```diff
@@ -119,6 +119,11 @@ Different flags (for the same filter target) combine with AND semantics. For
...
+
+Datacenter metrics can be aggregated to reduce the output size of the metrics
+endpoint by using the `-aggregate-only` flag.
```
leklund: I'm also not a wordsmith 😃 but how about this?

```suggestion
### Metrics Grouping: by datacenter or aggregate

The Fastly real-time stats API returns measurements grouped by datacenter
as well as aggregated measurements for all datacenters. 

By default, exported metrics are grouped by datacenter. The response
body size of the metrics endpoint can potentially be very large.
This will be exacerbated when using the exporter with many services,
many origins with Origin Inspector, and many domains with Domain Inspector. 

One way to reduce the output size of the metrics endpoint is by using
the `-aggregate-only` flag. When this flag is used only the `aggregated`
metrics from the real-time stats API will be exported. Metrics will still include
the datacenter label but it will always be set to "aggregate".
```

pkg/rt/subscriber.go
```diff
@@ -71,6 +72,13 @@ func WithPostprocess(f func()) SubscriberOption {
...
+// per datacenter metrics. By default, per datacenter are provided. Enabling
+// this feature can significantly reduce the payload of the metrics endpoint.
```
leklund: Looks good! I made a couple small tweaks.
```suggestion
// WithAggregateOnly sets whether aggregate metrics are output instead of
// per datacenter metrics. By default, per datacenter are provided. Enabling
// this feature will significantly reduce the payload size of the metrics endpoint.
```

leklund (approved):
A couple suggested wording changes to the docs but this looks good

README.md
mikelorant: Updated to your revision.

README.md
mikelorant: Updated to your revision.

pkg/rt/subscriber.go
mikelorant: Updated to your revision.

mikelorant:
Wording has been changed to your recommendations.

Been a long road to get this pull request to where we want it, but it seems we finally have something worthy of merging!

leklund:
@mikelorant Thanks for your patience and all the work you did in getting this merged! I'll cut a new release asap.

mikelorant:
Thanks so much for your efforts in getting this implemented. Been a great learning process and the solution implemented will really benefit users will large number of services.

PR #154, opened 2024-01-05
Update product cache to use /entitlements endpoint

dinoboy197:
Switch to use `/entitlements` endpoint rather than `/entitled-products`.

**Do not merge or create a release from this draft. This is for testing purposes only at the moment.**

### DIFF
--- a/pkg/api/product_cache.go
+++ b/pkg/api/product_cache.go
@@ -88,13 +88,7 @@ func (p *ProductCache) Refresh(ctx context.Context) error {
 	activeProducts := make(map[string]interface{})
 
 	for _, customer := range response.Customers {
-		if customer.Contracts == nil {
-			continue
-		}
 		for _, contract := range customer.Contracts {
-			if contract.Items == nil {
-				continue
-			}
 			for _, item := range contract.Items {
 				if item.ProductID == nil {
 					continue
--- a/pkg/api/product_cache.go
+++ b/pkg/api/product_cache.go
@@ -27,12 +27,20 @@ const (
 // Products is the slice of available products supported by real-time stats.
 var Products = []string{ProductDefault, ProductOriginInspector, ProductDomainInspector}
 
-// Product models the response from the Fastly Product Entitlement API.
-type Product struct {
-	HasAccess bool `json:"has_access"`
-	Product   struct {
-		Name string `json:"id"`
-	} `json:"product"`
+type response struct {
+	Customers *[]customer `json:"customers,omitempty"`
+}
+
+type customer struct {
+	Contracts *[]contract `json:"contracts,omitempty"`
+}
+
+type contract struct {
+	Items *[]item `json:"items,omitempty"`
+}
+
+type item struct {
+	ProductID *string `json:"product_id,omitempty"`
 }
 
 // ProductCache fetches product information from the Fastly Product Entitlement API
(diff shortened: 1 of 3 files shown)


pkg/api/product_cache.go
```diff
@@ -27,12 +27,20 @@ const (
...
+type customer struct {
+	Contracts *[]contract `json:"contracts,omitempty"`
+}
+
+type contract struct {
+	Items *[]item `json:"items,omitempty"`
+}
+
+type item struct {
+	ProductID *string `json:"product_id,omitempty"`
```
leklund: Since we only care about the productIDs deep in the nested slices and the other types aren't used anywhere else this could be simplified as a single type:

```
type entitlements struct {
	Customers []struct {
		Contracts []struct {
			Items []struct {
				ProductID *string `json:"product_id,omitempty"`
			} `json:"items,omitempty"`
		} `json:"contracts"`
	} `json:"customers"`
}
```

Also, no need to use `*[]` here.

pkg/api/product_cache.go
```diff
@@ -59,41 +67,60 @@ func NewProductCache(client HTTPClient, token string, logger log.Logger) *Produc
...
 
-		if resp.StatusCode != http.StatusOK {
-			return NewError(resp)
-		}
+	if resp.StatusCode != http.StatusOK {
+		return NewError(resp)
+	}
 
-		var response Product
+	var response response
```
leklund: This mainly style thing but I wouldn't recommend naming the variable the same as the type. Perhaps the type could be `entitlements` and this becomes

```
var response entitlements
```

pkg/api/product_cache_test.go
```diff
@@ -21,7 +21,7 @@ func TestProductCache(t *testing.T) {
 	}{
 		{
 			name:    "success",
-			client:  newSequentialResponseClient(productsResponseOne, productsResponseTwo),
+			client:  newSequentialResponseClient(productsResponse),
```
leklund: Since it's only a single request this could use the fixed response client:

```
fixedResponseClient{code: http.StatusOK, response: productsResponse}
```

(this test is the only place the sequential client is used but I can remove that in future cleanup PR)

leklund (commented):
A couple style comments but this is looking good. I also want to make sure the product names are correct (I think so but I want to check that with you).

pkg/api/product_cache.go
dinoboy197: Got it, thanks! Done.

pkg/api/product_cache.go
dinoboy197: Done!

pkg/api/product_cache_test.go
dinoboy197: Done, and I removed the sequential client that isn't in use after this.

pkg/api/product_cache.go
```diff
@@ -59,41 +61,58 @@ func NewProductCache(client HTTPClient, token string, logger log.Logger) *Produc
...
-		}
+	if err := json.NewDecoder(resp.Body).Decode(&response); err != nil {
+		return fmt.Errorf("error decoding API product response: %w", err)
+	}
+
+	activeProducts := make(map[string]interface{})
 
-		level.Debug(p.logger).Log("product", response.Product.Name, "hasAccess", response.HasAccess)
+	for _, customer := range response.Customers {
+		if customer.Contracts == nil {
```
leklund: You don't this `nil` check. You can range over a `nil` slice (it behave just like a slice with zero entries). Same for `contract.Items`

dinoboy197: Got it, thanks! Updated.

pkg/api/product_cache_test.go
leklund: Thank you!

leklund:
This looks great! Let me know when the new endpoint is ready to go and I can test this branch locally. Thanks for taking care of this!

PR #168, opened 2024-05-26
Fix aggregated domain metrics test fixture

mikelorant:
The aggregated metrics for the domain test had the host incorrectly set.

Fixed the indentation to be consistent with the other fixtures.

### DIFF
--- a/pkg/rt/common_test.go
+++ b/pkg/rt/common_test.go
@@ -4765,27 +4765,27 @@ const domainsResponseFixture = `{
 				}
 			},
 			"aggregated": {
-				"testorigin": {
-						"bandwidth": 1048,
-						"bereq_body_bytes": 16,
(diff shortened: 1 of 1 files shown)


mikelorant:
@leklund While implementing the tests for the domain metrics, I found that the domain was marked as an origin (likely copied from the origin metrics). This caused the tests to fail:

```text
 - 	`testspace_domain_bereq_body_bytes_total{datacenter="aggregate",domain="testhost",service_id="my-service-id",service_name="my-service-name"}`:                         16,
 + 	`testspace_domain_bereq_body_bytes_total{datacenter="aggregate",domain="testorigin",service_id="my-service-id",service_name="my-service-name"}`:                       16,
```

Since this is unrelated to the aggregated metrics, felt it was better to put this up as a separate pull request.

leklund (approved):

PR #169, opened 2024-05-26
Refactor metrics processing

mikelorant:
Setting metrics was tightly coupled with the gathering of the data. This made it difficult to change the logic without a significant amount of code changes.

By splitting the setting of the metrics into its own function, we can change the high level logic without having to move hundreds of lines around.


mikelorant:
@leklund Splitting out the `Process` refactor as its own pull request as this doesn't need to part of the aggregated datacenter changes.

leklund (approved):

PR #172, opened 2024-07-25
Fix a bucket mapping issue with the origin inspector latency metrics

crivera-fastly:
The exporter maps bucketed counters from the real-time stats API to a Prometheus histogram. Since it doesn't have the actual values that were used to build the bucketed counters it has to pick a value for each bucket to call Observe() with.

The current code uses values that are mapping to the wrong buckets. Here's a subset of the buckets from the origin latency histogram definition:

0.001, 0.005, 0.010

Calling Observe() with the value of 0.005 places the value in the 1-5ms bucket, not the 5-10ms bucket. The result of the current behavior is that the counts for each bucket are shifted to the next smallest bucket.

This fixes the issue by using the values at the end of each bucket interval as the parameters to Observe(). This will result in increased origin latency values when users upgrade to a release with this commit, but the increased numbers are more accurate.


leklund (approved):
Looks good :+1:

PR #202, opened 2025-06-06
Add an arm64 build

### LINKED ISSUE #200: Request: Support for the ARM architecture
Prometheus is available for both `amd64` and `arm64` architectures, so we should add cross-compilation so that the Exporter can be used in ARM envs as well.

case-fastly:
In addition to adding an `arm64` build, this tidies up the `ci.yaml` file a bit:

- Closes #200.
- Makes the `target` matrix more explicit
- Assigns names to some of the unnamed job steps, for quicker / easier debugging in the build output
- Adds some readability spacing

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,5 +1,7 @@
 name: CI
+
 on: [push, pull_request]
+
 jobs:
   test:
     strategy:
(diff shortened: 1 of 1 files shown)


cee-dub (commented):
Does the artifact need to take in any of the architecture variables?

cee-dub (approved):
I just missed that `ASSET_NAME` includes `GOARCH` and `GOOS`.

leklund (approved):

Israphel:
note: with this change, the docker container remains amd64 only.

PR #212, opened 2025-06-12
CI: add a Docker dry-run job

case-fastly:
There may be a better way to do this, but it seemed worth trying.

- Adds a new `docker-dry-run` job with `push: false` set, which means it will do the `docker build` but not push it to the GH package registry.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,8 +9,11 @@ jobs:
         platform: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.platform }}
     steps:
-      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


cee-dub (commented):
You might be able to use an artifact or output from the set run step to actually make the release, or that might add unnecessary coupling and it can repeat the work as it does here.

case-fastly:
> You might be able to use an artifact or output from the set run step to actually make the release, or that might add unnecessary coupling and it can repeat the work as it does here.

I think `ci.yaml` may actually need to be refactored, because it's currently doing a lot:

- Tests
- GH release
- Assets metadata for the GH release
- Docker dry-run
- Docker release

leklund (approved):

PR #213, opened 2025-06-18
update build tags to actually work with prometheus/common/version

leklund:
We added buildInfo metrics in #210  but didn't properly set the `ldflags` to populate the metrics.


case-fastly (approved):
Ship it.

PR #222, opened 2025-09-29
fix off-by-10x error in Compute timing metrics

joeshaw:
ComputeExecutionTimeMilliseconds, ComputeRequestTimeBilledMilliseconds,
and ComputeRequestTimeMilliseconds were all divided by 10,000 rather
than 1,000 for conversion to seconds.


cee-dub (approved):

joeshaw:
There are some obvious compatibility implications to this.  Perhaps this should be a completely new metric.  I have worked around this in my dashboards for now by multiplying the metric by 10 and including a comment referencing this bug and the versions that have it (9.5.0 and older).

leklund (approved):
well that's an unfortunate typo.

leklund:
> Perhaps this should be a completely new metric. I have worked around this in my dashboards for now by multiplying the metric by 10 and including a comment referencing this bug and the versions that have it (9.5.0 and older).

Instead of a new metric I think we can release this as 10.0.0 and consider this a "breaking change"?

joeshaw:
Yeah, that's reasonable too.  It just means that there'll be a point at which historical data is no longer compatible, but people probably aren't adjusting for this bug anyhow.

PR #224, opened 2025-10-22
added dictionary metrics to exporter

sflagg-fastly:


pkg/api/dictionaryinfo.go
```diff
@@ -0,0 +1,315 @@
...
+	req, err := http.NewRequestWithContext(ctx, method, url, nil)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Fastly-Key", c.token)
+	req.Header.Set("Accept", "application/json")
+	return req, nil
+}
+
+func (c *DictionaryInfoCache) listServices(ctx context.Context) ([]svc, error) {
```
crivera-fastly: I'm not super familiar with this codebase, but there's an existing `serviceCache` that's already refreshing the service list. Is there a reason that we can't use that in this file? That would avoid having two different caches refreshing the service list. I think using that would also respect the `service`, `service-allowlist`, `service-blocklist`, and `service-shard` flags.

sflagg-fastly: Makes sense. Thanks for pointing that out. I made the change.

pkg/api/dictionaryinfo.go
```diff
@@ -0,0 +1,338 @@
...
+	return items, json.NewDecoder(resp.Body).Decode(&items)
+}
+
+// Refresh queries Fastly APIs and rebuilds the in-memory snapshot.
+func (c *DictionaryInfoCache) Refresh(ctx context.Context) error {
+	if !c.enabled {
+		return nil
+	}
+	var services []svc
+	if c.serviceCache != nil {
```
leklund: `c.serviceCache` will never be nil

pkg/api/dictionaryinfo.go
```diff
@@ -0,0 +1,338 @@
...
+				},
+			})
+		}
+	} else {
+		var err error
+		services, err = c.listServices(ctx)
+		if err != nil {
+			return err
+		}
+	}
```
leklund: This else can be removed -- see above comment re: `c.serviceCache` which will always be not nil.

pkg/api/dictionaryinfo.go
```diff
@@ -0,0 +1,338 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode/100 != 2 {
+		return nil, fmt.Errorf("list services: %s", resp.Status)
+	}
+	var svcs []svc
+	return svcs, json.NewDecoder(resp.Body).Decode(&svcs)
+}
```
leklund: Should be removed since it will never be used.

pkg/api/dictionaryinfo.go
```diff
@@ -0,0 +1,338 @@
...
+
+type svc struct {
+	ID       string `json:"id"`
+	Name     string `json:"name"`
+	Versions []struct {
+		Number int  `json:"number"`
+		Active bool `json:"active"`
+	} `json:"versions"`
+}
```
leklund: I think this can be removed since the listServices can be removed and the suggested refactor to the Refresh method.

pkg/api/dictionaryinfo.go
```diff
@@ -0,0 +1,338 @@
...
+
+// Refresh queries Fastly APIs and rebuilds the in-memory snapshot.
+func (c *DictionaryInfoCache) Refresh(ctx context.Context) error {
+	if !c.enabled {
+		return nil
+	}
+	var services []svc
+	if c.serviceCache != nil {
+		for _, s := range c.serviceCache.Services() {
```
leklund: I think this range can be removed since it's just creating a `[]svc`.

pkg/api/dictionaryinfo.go
```diff
@@ -0,0 +1,338 @@
...
+	for _, layout := range layouts {
+		if t, err := time.Parse(layout, s); err == nil {
+			return float64(t.Unix()), true
+		}
+	}
+	return 0, false
+}
+
+// GET /service/{sid}/version/{v}/dictionary/{did}/item
+func (c *DictionaryInfoCache) listDictionaryItems(ctx context.Context, serviceID string, version int, dictID string) ([]dictItem, error) {
```
leklund: Dictionaries can have many thousands of items. This only appears to be used to get the last_updated timestamp. The only time that the dictionary info endpoint will return without last_updated is when there are no items in the dictionary. Because there is always a timestamp, this can be removed.

pkg/api/dictionaryinfo.go
```diff
@@ -0,0 +1,338 @@
...
+
+type DictionaryInfoCache struct {
+	client       *http.Client
+	token        string
+	logger       log.Logger
+	serviceCache *ServiceCache
+	enabled      bool
+
+	// Cached snapshot used by the collector to avoid network I/O on scrape.
+	snapshot []metricRow
```
leklund: I'd suggest renaming this to `dictionaries` or `dicts`

pkg/api/dictionaryinfo.go
```diff
@@ -0,0 +1,338 @@
...
+}
+
+type metricRow struct {
```
leklund: Suggested renaming: `dictionaryInfo`

pkg/api/dictionaryinfo.go
```diff
@@ -0,0 +1,338 @@
...
+		active := -1
+		for _, v := range s.Versions {
+			if v.Active {
+				active = v.Number
+				break
+			}
+		}
+		if active < 0 {
+			continue
+		}
```
leklund: Instead of using the `svc` structure. This can just iterate over the service cache:

```
for _, s := range c.serviceCache.Services() {
    if s.Version == 0 {
        continue
    }
    ... 
```

pkg/api/dictionaryinfo.go
```diff
@@ -0,0 +1,338 @@
...
+				if items, err := c.listDictionaryItems(ctx, s.ID, active, d.ID); err == nil && len(items) > 0 {
+					var maxTS float64
+					for _, it := range items {
+						if v2, ok2 := parseFastlyTime(it.UpdatedAt); ok2 && v2 > maxTS {
+							maxTS = v2
+						}
+					}
+					ts = maxTS // stays 0 if all missing/unparseable
+				}
+			}
```
leklund: This should be removed. Because dictionaries can have many thousand of times we don't want to iterate over them and request each item one-at-a-time.

if `info.LastUpdated` doesn't exist that means there are no items in the dictionary so we can just use the dictionary updated_at timestamp. The `dict` struct will need to have the updated_at field added.

pkg/api/dictionaryinfo.go
```diff
@@ -0,0 +1,338 @@
...
+		"2006-01-02T15:04:05",       // without zone
+		"2006-01-02 15:04:05Z07:00", // space + zone
+	}
+	for _, layout := range layouts {
+		if t, err := time.Parse(layout, s); err == nil {
+			return float64(t.Unix()), true
+		}
+	}
+	return 0, false
+}
```
leklund: This should be moved to the bottom and only need to support two layouts:

```
time.RFC3339,
"2006-01-02 15:04:05",
```

And because you don't need have the `bool` to fallback to listing dictionary items you can also remove the bool from the return signature.

pkg/api/dictionaryinfo.go
```diff
@@ -0,0 +1,338 @@
...
+	Name string `json:"name"`
+}
+
+// dictionary item model (subset)
+type dictItem struct {
+	ItemKey   string `json:"item_key"`
+	ItemValue string `json:"item_value"`
+	UpdatedAt string `json:"updated_at"`
+	CreatedAt string `json:"created_at"`
+}
```
leklund: can be deleted since we will longer be listing all items.

pkg/api/dictionaryinfo.go
```diff
@@ -0,0 +1,338 @@
...
+		prometheus.BuildFQName(namespace, subsystem, "dictionary_last_updated_timestamp_seconds"),
+		"Unix timestamp (seconds) when the dictionary was last updated (UTC).",
+		labelKeys, nil,
+	)
+
+	digestInfoDesc := prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, subsystem, "dictionary_digest_info"),
+		"Digest of the dictionary content (info metric; value is always 1).",
+		append(labelKeys, "digest"), nil,
+	)
```
leklund: Since dictionaries are versionless this could potentially be a unique digest on every update and will just lead to a cardinality explosion without providing. If this is absolutely necessary it should be added as an exemplar to the lastUpdated metric.

pkg/api/dictionaryinfo.go
```diff
@@ -0,0 +1,338 @@
...
+			r.ServiceID,
+			r.ServiceName,
+			strconv.Itoa(r.Version),
+			r.DictionaryID,
+			r.DictionaryName,
+		}
+		ch <- prometheus.MustNewConstMetric(dc.itemCountDesc, prometheus.GaugeValue, r.ItemCount, labels...)
+		ch <- prometheus.MustNewConstMetric(dc.lastUpdatedDesc, prometheus.GaugeValue, r.LastUpdatedTS, labels...)
+		digestLabels := append(labels, r.Digest)
+		ch <- prometheus.MustNewConstMetric(dc.digestInfoDesc, prometheus.GaugeValue, 1, digestLabels...)
```
leklund: Instead of a dedicated metric for the digest which will cause a cardinality explosion we could try adding it as an exemplar to the lastUpdated metric:

```
		m := prometheus.MustNewConstMetric(dc.lastUpdatedDesc, prometheus.GaugeValue, r.LastUpdatedTS, labels...)
		m = prometheus.MustNewMetricWithExemplars(m, prometheus.Exemplar{
			Value:     r.LastUpdatedTS,
			Labels:    prometheus.Labels{"digest": r.Digest},
		})
		ch <- m
```

I haven;t test this yet but it might work? 🤞

sflagg-fastly: I tried the exempler but got this
```
curl -s http://127.0.0.1:9416/metrics \
  | grep -E 'dictionary_(item_count|last_updated_timestamp_seconds|digest_info)'

* [from Gatherer #1] [from Gatherer #2] error collecting metric Desc{fqName: "fastly_rt_dictionary_last_updated_timestamp_seconds", help: "Unix timestamp (seconds) when the dictionary was last updated (UTC).", constLabels: {}, variableLabels: {service_id,service_name,version,dictionary_id,dictionary_name}}: cannot inject exemplar into Gauge, Summary or Untyped
* [from Gatherer #1] [from Gatherer #2] error collecting metric Desc{fqName: "fastly_rt_dictionary_last_updated_timestamp_seconds", help: "Unix timestamp (seconds) when the dictionary was last updated (UTC).", constLabels: {}, variableLabels: {service_id,service_name,version,dictionary_id,dictionary_name}}: cannot inject exemplar into Gauge, Summary or Untyped
```

sflagg-fastly: For now I’ve removed digest from the exported metrics entirely and only keep it in the internal snapshot.

pkg/api/dictionaryinfo.go
sflagg-fastly: I've made all the suggested changes. Thanks for your review!

pkg/api/dictionaryinfo.go
```diff
@@ -0,0 +1,229 @@
...
+				Version:        active,
+				DictionaryID:   d.ID,
+				DictionaryName: d.Name,
+				Digest:         info.Digest,
+				ItemCount:      float64(info.ItemCount),
+				LastUpdatedTS:  ts,
+			})
+		}
+	}
+	c.snapshot = out
```
leklund: You'll want to wrap this in a mutex (and add a `sync.RWMutex` to the `DictionaryInfoCache` struct)

```
c.mutex.Lock
c.snapshot = out
c.mtx.Unlock()
```

pkg/api/dictionaryinfo.go
```diff
@@ -0,0 +1,229 @@
...
+	lastUpdatedDesc *prometheus.Desc
+}
+
+func (dc *dictionaryInfoCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- dc.itemCountDesc
+	ch <- dc.lastUpdatedDesc
+}
+
+func (dc *dictionaryInfoCollector) Collect(ch chan<- prometheus.Metric) {
+	for _, r := range dc.cache.snapshot {
```
leklund: Instead of using the snapshot directly it's probably better to create a new method to return a snapshot that can be wrapped in a mutex:

```
func (c *DictionaryInfoCache) Dictionaries() []Dictionary {
    c.mtx.Lock()
    defer c.mtx.Unlock()
    dictionaries := c.dictionaries[:]
    return dictionaries
```

This assumes that `metricRow` is renamed to `Dictionary` and `snapshot` is renamed to `dictionaries`

pkg/api/dictionaryinfo.go
```diff
@@ -0,0 +1,229 @@
...
+	"strconv"
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// DictionaryInfo is the Fastly API response for the /info endpoint.
+type DictionaryInfo struct {
```
leklund:
```suggestion
// dictionaryInfo is the Fastly API response for the /info endpoint.
type dictionaryInfo struct {
```

The struct that represents the API response doesn't need to be exported

pkg/api/dictionaryinfo.go
```diff
@@ -0,0 +1,229 @@
...
+	Digest      string `json:"digest"`
+	ItemCount   int64  `json:"item_count"`
+	LastUpdated string `json:"last_updated"` // may be RFC3339 or "2006-01-02 15:04:05"; can be empty
+}
+
+type dict struct {
```
leklund: Took me a minute to figure out what this was for. I'd rename this to dictionaryResp or something desscriptive and then add a comment

```
# dictionaryResp represents an individual dictionary from the listDictionaries API call
```

pkg/api/dictionaryinfo.go
leklund: Or maybe just `Dictionary`

leklund (commented):
This is looking good! I added a few more comments. We need a mutex around the snapshot slice and I added some renaming recommendations

sflagg-fastly:
I've made all the changes. How does this look now?

leklund:
@sflagg-fastly Thanks! This looks great! I added a test and made a few tweaks to make the DictionaryCache more like the other Caches.

leklund (approved):

PR #246, opened 2026-05-12
Make Fastly API and RT base URLs configurable via envvars

grantstephens:
## Summary

Add support for a single `FASTLY_BASE_URL` environment variable to override the default Fastly base domain (`fastly.com`). The API and RT URLs are derived as subdomains:

- API: `https://api.{FASTLY_BASE_URL}`
- RT:  `https://rt.{FASTLY_BASE_URL}`

For example, setting `FASTLY_BASE_URL=staging.fastly.net` would route API calls to `https://api.staging.fastly.net` and real-time stats to `https://rt.staging.fastly.net`.

When unset, the existing defaults (`fastly.com`) are used — fully backward compatible.

## Motivation

Useful for testing against staging environments, local development proxies, or alternative API endpoints without modifying code.

## Changes

- Added `DefaultBaseDomain` constant and `APIBaseURL()`/`RTBaseURL()` helper functions in `pkg/api/common.go`
- Added `baseURL` field to all API cache structs and the RT subscriber
- Added `rt.WithBaseURL()` subscriber option
- Updated constructors to accept a `baseURL` parameter (empty string = default)
- Read `FASTLY_BASE_URL` in `main.go` and derive both URLs from it
- Updated all tests to pass the new parameter

### DIFF
--- a/cmd/fastly-exporter/main.go
+++ b/cmd/fastly-exporter/main.go
@@ -80,7 +80,8 @@ func main() {
 		fs.BoolVar(&configFileExample, "config-file-example", false, "print example config file to stdout and exit")
 		fs.Usage = usageFor(fs)
 	}
-	if err := ff.Parse(fs, os.Args[1:],
+	if err := ff.Parse(
+		fs, os.Args[1:],
 		ff.WithEnvVarPrefix("FASTLY_EXPORTER"),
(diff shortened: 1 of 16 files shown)


awilliams-fastly:
What was the reason for closing this PR?

